PR #17, opened 2022-12-15
deps: updating dependencies owing to vulnerabilities

laserkelvin:
This commit changes the requirements for both PyTorch and `pymatgen` to avoid vulnerabilities in the pinned versions.


laserkelvin:
Addressed in #16 and #15 from dependabot PRs

PR #49, opened 2023-09-07
Removing Hard Coded Path in Nomad Example

melo-gonzo:
DataModule should be using `.from_devset()`. Updating example to reflect this.


laserkelvin (approved):
Looks good :)

PR #56, opened 2023-10-18
Adding M3GNet from matgl 

melo-gonzo:
This PR adds in the matgl repository, which implements the M3GNet model. The default model from matgl needed modification in its forward pass, which is overwritten in the matsciml model file (`./matsciml/models/dgl/m3gnet.py`). Additional modifications may be needed to output only the graph embedding prior to the final layer - this may be added as a feature to matgl itself to prevent the need to copy over code. Two tests are added to check importing the model, and a simple forward pass which checks for output dimensions and gradients. 

The input graph's are also expected to have a different structure, as well as a few more node and edge features. The processing step of matgl ([here](https://github.com/materialsvirtuallab/matgl/blob/0207660e41f7c05c412331541e8ebb98d8c9c1e9/matgl/graph/data.py#L252)) is implemented and a new dataset subclass of `MaterialsProjectDataset` is used to incorporate this processing and additional hyperparameters.

Finally, the matgl package is added as a requirement.

------------------
After addressing some comments, the subclass datasets for M3GNet are added to their respective `dataset.py` file. Additionally, the dictionary mapping atomic symbols to numbers was added to `matsciml/datasets/utils.py` file, as it is reused often. Constants were added into `NomadDataset` to convert the units to the format used by all other datasets (i.e., from Joules to eV, and from meters to angstroms). 

M3GNet is now able to run on Materials Project, Nomad, OQMD, and Carolina Materials database.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,8 +6,7 @@ logs
 *.traj
 experimental
 lightning_logs
-*.lmdb
-*.lmdb-lock
+
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
--- a/.gitignore
+++ b/.gitignore
@@ -6,9 +6,8 @@ logs
 *.traj
 experimental
 lightning_logs
-*commands*
-*kdp*
-
+*.lmdb
+*.lmdb-lock
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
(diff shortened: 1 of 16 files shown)


smiret-intel:
Let's supplement the examples with additional datasets (e.g. OQMD, NOMAD, CMD)

melo-gonzo:
@laserkelvin @smiret-intel please see the updates and re-review.

smiret-intel (approved):

laserkelvin (approved):

laserkelvin:
LGTM

PR #60, opened 2023-10-23
Pre-commit fix and adding `isort`

laserkelvin:
This PR adds a working version of `isort` into the `pre-commit` workflow, in addition to patching up the `black` specification to something more recent. Both are verified to work on my laptop environment.


melo-gonzo (approved):
looks good 👍

melo-gonzo (approved):

PR #66, opened 2023-11-17
Adding PyG Version of FAENet

melo-gonzo:
This PR adds the FAENet architecture, and frame averaging transform, which may be used with other architectures. Additionally, a transform for calculating a material's unit cell from lattice parameters is added, which is needed to generate extra keys to use FAENet. 

Original FAENet implementation: https://github.com/vict0rsch/faenet

### DIFF
--- a/matsciml/common/utils.py
+++ b/matsciml/common/utils.py
@@ -4,6 +4,7 @@ Copyright (c) Facebook, Inc. and its affiliates.
 This source code is licensed under the MIT license found in the
 LICENSE file in the root directory of this source tree.
 """
+from __future__ import annotations
 
 import ast
 import collections
@@ -25,16 +26,19 @@ import torch
 import yaml
 from matplotlib.backends.backend_agg import FigureCanvasAgg as FigureCanvas
 from matplotlib.figure import Figure
+from torch_scatter import segment_coo, segment_csr
 
 
 def save_checkpoint(
-    state, checkpoint_dir="checkpoints/", checkpoint_file="checkpoint.pt"
+    state,
+    checkpoint_dir="checkpoints/",
+    checkpoint_file="checkpoint.pt",
 ):
     filename = os.path.join(checkpoint_dir, checkpoint_file)
     torch.save(state, filename)
 
 
-class Complete(object):
+class Complete:
     def __call__(self, data):
         device = data.edge_index.device
 
@@ -73,7 +77,7 @@ def warmup_lr_lambda(current_step, optim_config):
         or "warmup_epochs" in optim_config
     ):
         raise Exception(
-            "ConfigError: please define lr_milestones in steps not epochs and define warmup_steps instead of warmup_epochs"
+            "ConfigError: please define lr_milestones in steps not epochs and define warmup_steps instead of warmup_epochs",
         )
 
     if current_step <= optim_config["warmup_steps"]:
@@ -132,7 +136,7 @@ def plot_histogram(data, xlabel="", ylabel="", title=""):
     canvas.draw()
     image_from_plot = np.frombuffer(fig.canvas.tostring_rgb(), dtype=np.uint8)
     image_from_plot = image_from_plot.reshape(
-        fig.canvas.get_width_height()[::-1] + (3,)
+        fig.canvas.get_width_height()[::-1] + (3,),
     )
 
     return image_from_plot
@@ -165,7 +169,8 @@ def collate(data_list):
     for key in keys:
         if torch.is_tensor(data_list[0][key]):
             data[key] = torch.cat(
-                data[key], dim=data.__cat_dim__(key, data_list[0][key])
+                data[key],
+                dim=data.__cat_dim__(key, data_list[0][key]),
             )
         else:
             data[key] = torch.tensor(data[key])
@@ -243,7 +248,7 @@ def setup_imports():
                 splits = f.split(os.sep)
                 file_name = splits[-1]
                 module_name = file_name[: file_name.find(".py")]
-                importlib.import_module("matsciml.%s.%s" % (key[1:], module_name))
+                importlib.import_module(f"matsciml.{key[1:]}.{module_name}")
 
     experimental_folder = os.path.join(root_folder, "../experimental/")
     if os.path.exists(experimental_folder):
@@ -258,7 +263,8 @@ def setup_imports():
             with open(ignore_file) as f:
                 for line in f.read().splitlines():
                     ignored += glob.glob(
-                        experimental_folder + line + "/**/*py", recursive=True
+                        experimental_folder + line + "/**/*py",
+                        recursive=True,
                     )
             for f in ignored:
                 experimental_files.remove(f)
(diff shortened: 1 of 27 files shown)


matsciml/common/utils.py
laserkelvin, line 1: What is all this??

laserkelvin: Oh, did `black` just run on this file for the first time?

matsciml/datasets/transforms/props.py
```diff
@@ -407,6 +416,24 @@ def __call__(self, data: DataDict) -> DataDict:
...
+        lattice_features = data.get("lattice_features", None)
+        if lattice_features is None:
+            lattice_params = data.get("lattice_params", None)
+        else:
+            lattice_params = lattice_features.get("lattice_params", None)
+        if lattice_params is None:
+            raise KeyError(
+                f"No lattice parameters available to calculate unit cell matrix.",
+            )
+        abc, angles = lattice_params[:3], lattice_params[3:]
```
laserkelvin, line 431: Do you need to convert angles back to degrees for `Lattice`?

matsciml/common/utils.py
melo-gonzo: Yeah I think its from black or one of the pre-commit hooks. I just had to update an import here, `from torch_scatter import segment_coo, segment_csr`, and then things got reformatted like crazy. Happy to revert that if you'd like.

matsciml/models/common.py
laserkelvin, line 1: Did you edit anything in this file that isn't from the hooks?

matsciml/models/pyg/faenet/helper.py
```diff
@@ -0,0 +1,156 @@
...
+import random
+
+import torch
+import torch.nn as nn
+import torch_geometric
+from torch_geometric.nn import radius_graph
+from torch_geometric.transforms import LinearTransformation
+
+
+def swish(x):
```
laserkelvin, line 14: Is this really necessarily LOL

matsciml/models/pyg/faenet/helper.py
```diff
@@ -0,0 +1,156 @@
...
+    """Swish activation function"""
+    return torch.nn.functional.silu(x)
+
+
+def get_pbc_distances(
+    pos,
```
laserkelvin, line 20: I'll help with this, but just wanted to leave a TODO to add type annotations (even though there are comments)

matsciml/common/utils.py
laserkelvin: I think we're through the looking glass: since you knew exactly what you've done I'd say we keep it. The pain is just separating actual code changes from the pre-commit stuff, so in the future I think we should run the hooks on every file in one big PR but until then I run the pre-commit hooks to check for issues but I don't add the style changes to the commit if they're not relevant.

matsciml/models/common.py
melo-gonzo: I did actually, but I'm realizing now that my change is unnecessary - I will revert this file back to original.

matsciml/models/pyg/faenet/helper.py
melo-gonzo: No comment here other than I just copied this part in from [original implementation ].(https://github.com/vict0rsch/faenet/blob/1f725cfabb0ef47a44662eac7bc39325249226b2/faenet/utils.py#L12)

laserkelvin: I don't think it's used anywhere, can we delete it?

melo-gonzo: [It is used](https://github.com/melo-gonzo/matsciml/blob/3a0852afdd76c34f72113751a55b17a4a25a9729/matsciml/models/pyg/faenet/faenet.py#L149), i guess people usually want to use "swish" for "silu" becasue it sounds cooler? Not sure, i think we should just leave it in for now.

matsciml/datasets/transforms/props.py
melo-gonzo: Good catch, fixed this. Assumes those lattice params angles are in radians though. Not sure if that is true for all datasets.

matsciml/models/pyg/faenet/helper.py
laserkelvin: Not a hill I'll die on (and you don't have to do it) but IMO if this is the kind of configuration we want to use, then we should set it up via the `registry`

melo-gonzo: Purged out the swish ✅

matsciml/models/pyg/faenet/faenet.py
```diff
@@ -0,0 +1,452 @@
...
+from matsciml.common.types import BatchDict
+from matsciml.common.types import DataDict
+from matsciml.common.types import Embeddings
+from matsciml.common.utils import radius_graph_pbc
+from matsciml.models.base import AbstractPyGModel
+from matsciml.models.pyg.faenet.helper import *
+from matsciml.models.pyg.faenet.layers import *
+
+
+class FAENet(AbstractPyGModel):
```
laserkelvin, line 24: Can you add this to the model registry?

```python
from matsciml.common.registry import registry
...

@registry.register_model("FAENet")
class FAENet(AbstractPyGModel):
    ...
```

matsciml/models/pyg/faenet/faenet.py
```diff
@@ -0,0 +1,452 @@
...
+
+        # check that frame averaging properties are available
+        for key in ["fa_pos", "fa_cell", "fa_rot"]:
+            if not hasattr(batch, key):
+                raise KeyError(
+                    f"Graph is expected to have property {key}: include frame averaging transform!",
+                )
+
+        # Distinguish Frame Averaging prediction from traditional case.
+        if frame_averaging and frame_averaging != "DA":
```
laserkelvin, line 406: Because of L391, it will never be "DA" right?

matsciml/models/pyg/faenet/faenet.py
```diff
@@ -0,0 +1,452 @@
...
+
+            # Compute model prediction for each frame
+            all_embeddings = []
+            for frame_idx, frame in enumerate(batch.fa_pos):
+                # set positions to current frame
+                batch.pos = frame
+                if crystal_task:
+                    batch.cell = batch.fa_cell[frame_idx]
+                # Forward pass
+                embeddings = self.first_forward(deepcopy(batch))
```
laserkelvin, line 418: I left this in, but do we actually need to do a `deepcopy`? Somewhat inefficient

laserkelvin (changes requested):

matsciml/common/utils.py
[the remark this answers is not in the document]
melo-gonzo: i believe i added in this import: `from torch_scatter import segment_coo, segment_csr`. other than that it was from pre-commit.

matsciml/models/pyg/faenet/faenet.py
melo-gonzo: it can be DA if that is what the transform uses. originally i hard coded that variable for simplicity. just changed up the model init args to take `frame_averaging` as an argument to make it consistent.

matsciml/models/pyg/faenet/faenet.py
melo-gonzo: deepcopy doesn't seem to do anything useful anymore.. i removed it.

laserkelvin (approved):
LGTM, please merge when you're ready!

PR #67, opened 2023-11-17
Model output refactor

laserkelvin:
This PR adds a bit of flexibility into task development by having a standardized data structure for what models return.

We introduce the `matsciml.common.types.Embeddings` dataclass, which holds `system_embeddings` and `point_embeddings` optional attributes, which house (as the name suggests) system/graph and point/node level embeddings from a model.

Prior to this, models would only return the system/graph-level embedding, which are then used by the task output heads for projection/prediction. In some cases, tasks will would want to operate on the point/node-level embeddings, and so this refactor facilitates that.

Most common models have been refactored to emit this `Embeddings` structure, and I've checked that they still run in the same way as they did before. In `AbstractTask`, which is what most model implementations should be inheriting from, the `forward` method will also check to make sure the output is an `Embeddings` instance to help flag down things that haven't been ported over.


melo-gonzo (approved):
Looks good. We should probably remove the `encoder_only` arg throughout all the models at some point as well so there isn't any confusion with what is expected to be returned from the forward call.

PR #73, opened 2023-12-01
Pre-commit update

laserkelvin:
This PR fixes and updates the `pre-commit` specification:

- `language_version` key for `black` is changed from `python3.9` to `python`, which resolves venv issues when it is not found.
- `black` is bumped to `23.7.0`
- `bandit` is a new part of the workflow, which automates scanning for unsafe code practices


melo-gonzo (approved):
looksGTM!

PR #72, opened 2023-11-28
Adding MACE model and equivariant output block

Vaibhav525:
🚀 Feature request
To add MACE pyg model and an equivariant output block. 
Request attributes
      - Would this be a refactor of existing code? - Yes 
      - Does this proposal require new package dependencies? - Yes e3nn
      - Would this change break backwards compatibility? -No
      - Does this proposal include a new model? -Yes , MACE
      - Does this proposal include a new dataset? -No
      - Does this proposal include a new task/workflow? Yes, equivariant output block

PR #74, opened 2023-12-07
Preserving symmetry representations in outputs

### LINKED ISSUE #70: [Feature request]: Preserving equivariance in model outputs
### Feature/behavior summary

In the current workflow, every model is expected to emit a pair of tensors for embeddings that represent points/nodes and systems/graphs. With models like those that rely on tensor products, and/or those that learn multiple types or components of embeddings per hierarchical level, reducing prematurely just to get embeddings that conform to those forms will lose information (i.e. even if we flatten).

Implementation-wise, this is the difference:

```python
...
# model architecture
...
return {"node": node_embeddings, "graph": graph_embeddings}
```

We want:

```python
...
# tensor product architecture
...
return {
    "node": {"irrep_0": irrep_0_node, "irrep_1": irrep_1_node},
     "graph": {"irrep_0": irrep_0_graph, "irrep_1": irrep_1_graph},
}
```

### Request attributes

- [X] Does this proposal include a new task/workflow?

### Related issues

_No response_

### Solution description

Currently, tasks simply use `OutputBlock` as projection layers that operate on the embeddings returned. The two things we need to do to support this are:

1. Implement an `OutputBlock` class that supports a hierarchy of embeddings, e.g. `e3nn` `Irreps`
2. Refactor tasks to support the usage of this new projection set

### Additional notes

_No response_

laserkelvin:
This PR implements support for `e3nn` features in support of #72. The main change is addition of `IrrepOutputBlock` that utilizes `e3nn.o3.Linear` instead of the regular PyTorch `Linear` to preserve symmetry representations. The `OutputHead` class was refactored to support use of either `IrrepOutputBlock` or the regular `OutputBlock`.

A simple example for using this workflow is forthcoming, but in the meantime, the API remains pretty consistent between the two workflows via `output_kwargs` in task specifications:

```python
task = ScalarRegressionTask(
   ...,
   output_kwargs={
      "block_type": "IrrepOutputBlock",
      "input_dim": "10x0e + 20x1e", 
      "hidden_dim": "30x0e + 10x1e", 
      "output_dim": "0e",
      "activation": ["torch.nn.SiLU", None]
    }
)
```

Closes #70.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -60,3 +60,4 @@ repos:
         language_version: python3
         types: [python]
         require_serial: true
+        args: [--skip, "B101", --recursive]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -36,6 +36,7 @@ repos:
           flake8-use-fstring,
           flake8-pyproject,
         ]
+        args: [--max-line-length=88, --ignore=S101]
 -   repo: https://github.com/psf/black
     rev: 23.7.0
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -37,15 +37,15 @@ repos:
           flake8-pyproject,
         ]
 -   repo: https://github.com/psf/black
-    rev: 22.3.1
+    rev: 23.7.0
     hooks:
       - id: black
-        language_version: python3.9
+        language_version: python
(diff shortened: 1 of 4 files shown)


laserkelvin:
> LgoodTM! do you want to add e3nn to the toml file in this pr or let the other one handle that?

Woops, thanks for that! I thought I already did

PR #76, opened 2023-12-08
Trigger `RuntimeError` for when normalization is requested but not utilized

### LINKED ISSUE #75: [Bug]: Normalization keys mismatch fails silently
### Expected behavior

When normalization parameters are passed, if there are no keys matched up we should throw an error or warning message to notify the user that nothing is being normalized.

### Actual behavior

Currently, there is no "validation" persay of normalization keys passed, and so if they aren't matched with any of the targets, nothing happens.

Pass normalization kwargs, and have none of the keys match available targets.

### Specifications

`matsciml` as of 46dd595

laserkelvin:
This PR adds checks and informative error messages centered around target normalization.

1. A check to `BaseTaskModule._compute_losses`, appending keys to a `used_norm` list as normalization is performed. When all the losses have been calculated, if `used_norm` is empty despite having requested normalization, we trigger a `RuntimeError` to stop training, and inform the user that no normalization was performed.
2. Another subsequent check to make sure that the number of keys used matches the number of keys specified by the user.

@melo-gonzo I've made it trigger an error, rather than a warning message, since it's probably too easy to ignore warnings and rather have it complete training.

Closes #75

### DIFF
--- a/matsciml/models/base.py
+++ b/matsciml/models/base.py
@@ -909,6 +909,11 @@ class BaseTaskModule(pl.LightningModule):
                 "Target normalization was intended but not used."
                 f"Please check your config - expected: {self.task_keys}"
             )
+        if len(used_norm) != len(self.normalizers):
+            raise RuntimeError(
+                "Normalization was performed, but number of keys do not match."
+                f"Expected {len(self.normalizers)} keys, but only used {len(used_norm)}."
+            )
         return {"loss": total_loss, "log": losses}
 
     def configure_optimizers(self) -> torch.optim.AdamW:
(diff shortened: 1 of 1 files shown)


melo-gonzo:
I like the idea, but I'm not sure this actually does anything at the moment. The [_make_normalizers](https://github.com/melo-gonzo/matsciml/blob/83d7afa4b9b4b3ca55fd08b103da027878289cbe/matsciml/models/base.py#L998) function takes in whatever `normalize_kwargs` are passed, but if they do not contain any of the task keys, it just writes out default values to the expected task keys normalizer and omits whatever was passed in.

PR #83, opened 2024-01-08
Implementing a MACE transform for expected data structures

laserkelvin:
This PR contributes to #72 by refactoring the `read_batch` functionality in the MACE model definition into its own transform, to keep things modular and more maintainable. This will require a number of things to be changed in the current implementation, so that `read_batch` actually is just retrieving and mapping the right keys as opposed to implementing some data logic. One of the main differences is that it packs all of the expected data into the `PyGGraph` structure, because the transform executes _before_ batching happens.

The first part of the function basically just checks for some tensors under specific names, and if they're not there, create dummy ones. The second bit is atom featurization: we take one-hot encodings from `torch.eye`, as opposed to goig through the whole `AtomicNumberTable` and then converting.

### DIFF
--- a/matsciml/datasets/transforms/mace.py
+++ b/matsciml/datasets/transforms/mace.py
@@ -1,0 +1,63 @@
+from __future__ import annotations
+
+import torch
+
+from matsciml.common.types import DataDict, PyGGraph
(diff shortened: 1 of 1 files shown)


laserkelvin:
With #156 in particular, this is does not seem to be needed anymore.

PR #86, opened 2024-01-09
82 feature request allow user specified reductions for embeddings

melo-gonzo:
Added kwarg to `BaseTaskModule` to set the embedding reduction type. Defaults to mean.


laserkelvin (approved):
LGTM, thanks for implementing this!

PR #89, opened 2024-01-10
Documentation on how to implement new datasets

### LINKED ISSUE #85: Example on how to create your own lmdb dataset for training.
### Feature/behavior summary

We have some larger material datasets we would like to train on and your repo is the best I could find in terms of support for large-scale training.
It would be great if you could provide an example for how to create your own lmdb dataset similar to the existing ones from a list of structures and properties and use that for training. I am sure it's quite simple but I have to admit I am getting a bit lost in all the different  classes of the datasets.
Or maybe you can point me in the right direction, i.e. what key and dictionary structure would be correct to use as input  to `write_lmdb_data(key: Any, data: Any, target_lmdb: lmdb.Environment)`  to be consistent with the existing datasets e.g. the `SinglePointLmdbDataset`.

thank you so much for your help.

### Request attributes

- [ ] Would this be a refactor of existing code?

### Related issues

_No response_

### Solution description

Add an example on how to create your own lmdb dataset from a list of structures and properties to the tutorials.

### Additional notes

_No response_

laserkelvin:
In response to #85, I wrote up a brain dump of how to use the `matsciml` API to implement new datasets from start to finish.

I suggest reading it from [this link](https://github.com/laserkelvin/matsciml/blob/85-dataset-documentation/matsciml/datasets/README.md) as opposed to reading it from the diffs/commits :)

Would be nice to receive some feedback and contributions from @melo-gonzo and @bmuaz, but also from @JonathanSchmidt1 to see if anything is missing or requires more elaboration, etc. For the purposes of merging, however, I only need @melo-gonzo to approve.

Closes #85.

### DIFF
--- a/matsciml/datasets/README.md
+++ b/matsciml/datasets/README.md
@@ -30,6 +30,7 @@ The following is a checklist to help guide you through the implementation of a n
 - [ ] Created new dataset interface by subclassing `matsciml.datasets.base.BaseLMDBDataset`
 - [ ] Implemented unit tests for new dataset class
 - [ ] Test in minimal training loop
+- [ ] Add your dataset import to `matsciml/datasets/__init__.py`
 
 If you think your implementation will benefit the community (hint: almost definitely), we encourage
 you to add the following steps as well:
--- a/matsciml/datasets/README.md
+++ b/matsciml/datasets/README.md
@@ -250,6 +250,34 @@ complicated quickly, so for pull requests, the bar is likely not going to be too
 these dataset unit tests. At a minimum, we ask you to replicate the tests above for your
 dataset, but highly encourage more expansive testing (you can never have too much!).
 
+### Final integration
+
+The last part of integrating your new dataset is using it to train a model! From this
+point onwards it depends entirely on what you want to do with your data, and so we only
+provide a very minimal example.
+
+```python
+import pytorch_lightning as pl
+
+from matsciml.datasets import new_materials_data
+from matsciml.lightning.data_utils import MatSciMLDataModule
+
+# this uses the devset for training
+data_module = MatSciMLDataModule.from_devset(
+    dataset="NewMaterialsDataset",
+    batch_size=8
+)
+# configure your task
+task = ...
+# fast_dev_run performs a minimal training loop with 10 steps
+# without checkpointing, logging, etc.
+trainer = pl.Trainer(fast_dev_run=10)
+trainer.fit(task, datamodule=data_module)
+```
+
+With this, you should be able to test the end-to-end functionality of your dataset;
+here's hoping everything works!
+
 ## Inheritance
 
 ## Common key names for data properties
--- a/matsciml/datasets/README.md
+++ b/matsciml/datasets/README.md
@@ -1,0 +1,66 @@
+
+# Dataset implementation in Open MatSciML Toolkit
+
+This document serves as a short overview of the dataset portion of the Open MatSciML Toolkit pipeline is designed,
+both for the sake of maintenance as well as to serve as a reference for subsequent dataset implementations.
+
+## Design philosophy
+
+- Inheritance/object-oriented programming is used to maximize code reuse.
+- Dataset methods should operate on *single* data samples, and are expected to return a `DataDict`, which is simply a dictionary consisting of string keys and various data formats (e.g. `float`, `torch.Tensor`).
+- If possible, return a point cloud sample as it does not require specific graph backends like PyG or DGL. Transforms are written to convert between representations.
+- Implement the *bare minimum* logic for `data_from_key` and use *transforms* to achieve modular behavior (e.g. convert point clouds to graphs, add optional properties).
+- When possible, use [consistent keys for properties](#common-key-names-for-data-properties).
+- Nest training targets under the `targets` key of your `DataDict`. This makes it explicit which tensors are intended to be used as outputs rather than input data, both for the pipeline as well as other users and maintainers.
+  - The `target_keys` property categorizes targets into continuous (`regression`) and binary (`classification`) keys, which are subsequently used by the task abstractions.
+- Commit a "development set" amount of data samples to the repository. This is typically around ~200 samples, and provides a straightforward way to test various stages of the pipeline, develop models offline, etc.
+- Develop unit tests as you go, testing individual dataset functionality as you develop it.
+
+## Implementing a new dataset
+
+The first step in contributing a new dataset is reading the [contribution guide](../../CONTRIBUTING.md) to configure your environment and such. At a high level,
+fork this repository, clone your fork locally, then create a new branch for code implementations. Assuming you have done this, the first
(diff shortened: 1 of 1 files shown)


JonathanSchmidt1:
This looks great, I will try to go through it tomorrow  and see if I have any questions.  Thank you so much already :)

melo-gonzo:
This is looking good! Thanks for writing it up. I would recommend adding a few more items.
* Note to add the dataset and available target keys to the table [here](https://github.com/laserkelvin/matsciml/blob/85-dataset-documentation/matsciml/datasets/DATASETS.md)
* Note to add the dataset to `matsciml/dataset/__init__.py`
* Note clarifying that dataset needs to be added to the registry

Do we need to add anything about ensuring licenses are compatible?

laserkelvin:
Thanks @melo-gonzo for the suggestions!

I've added commits for each suggestion. I've included a bit on the original dataset licensing, but whether or not they will be compatible or not probably should be determined in a offline discussion between us (the maintainers) and whoever submitting the PR, given the many complicated implications around dataset licensing.

My initial thought is that this is primarily going to be relevant to Open MatSciML Toolkit only in terms of `devset` distribution: it'll be up to the end-user to determine whether their application (academic/commercial) is compatible with Open MatSciML Toolkit's license and the license of datasets they are training on. (IANAL)

bmuaz:
Good thoughts. I agree with @laserkelvin. For us, we don't provide the whole datasets, 100 or 200 datapoints in devsets are not something people can use for the commercial applications. It will be the user's responsibility to check or negotiate with the datasets owners for their end-use, IMO.

melo-gonzo (approved):
This all looks good to me with the additional changes.

laserkelvin:
Thanks! I'll just wait and see if @JonathanSchmidt1 has any feedback before merging.

JonathanSchmidt1:
I managed to get our dataset working with the instructions so the documentation is already working ;). Thank you very much.
The only thing that was not entirely clear is how node level targets should be added. There is the warning about having force as a key but I would expect to have forces under 'targets' . Maybe you can add an explanation in case per atom targets are treated differently than per system targets. Or maybe I just missed something.
Ps: I was implementing the dataset class for the alexandria database with around 5M DFT  relaxed 3D, 2D and 1D crystal structures right now (in the process of being published in this updated version). Would you be interested in having it added to the package?

laserkelvin:
> The only thing that was not entirely clear is how node level targets should be added. There is the warning about having force as a key but I would expect to have forces under 'targets' .

I have a vague notion where that warning is being raised, but to be sure do you mind sharing the warning message?

> Maybe you can add an explanation in case per atom targets are treated differently than per system targets. Or maybe I just missed something.

I don't know if this will help or make things worse, but at the `Dataset` level (and indeed at the PyTorch level) all are tensors, and there's no real delineation of node/point, edge, or graph/cloud level features and labels. That structure is imposed by how you want to represent the data (either as point clouds or graphs): the loss calculation at the end of the day just cares that the tensor dimensions match.

That said, forces are kind of a weird thing: DGL differentiates between node/edge attributes whereas PyG does not, and so for correct batching and whatnot we do add forces to `DGLGraph.ndata`. If I remember correctly, we basically create a "link" (i.e. not a hard copy) between the `ndata` forces with the one residing in `targets`, so that when it comes to using the forces as labels the training pipeline has the same behavior regardless.

I'll try and add a bit of "well formed" explanations in the documentation. Initially I was going to say "We don't do anything differently", but then you got me thinking about forces and other potential node-level targets and whatnot. This might take a little more thought to standardize...

> Ps: I was implementing the dataset class for the alexandria database with around 5M DFT relaxed 3D, 2D and 1D crystal structures right now (in the process of being published in this updated version). Would you be interested in having it added to the package?

That sounds great! If you can share any more details on the dataset ahead of time, want to discuss implementation details, etc before submitting a PR, please feel free to set up a [discussion](https://github.com/IntelLabs/matsciml/discussions/categories/new-datasets)!

JonathanSchmidt1:
The warning was actually raised when running some tests include the dgl graph transform.
```
matsciml/datasets/alexandria/tests/test_alexandria.py::test_graph_transform
  /home/sjonathan/Downloads/matsciml_alexandria/matsciml/datasets/transforms/representations/graphs.py:166:
UserWarning: Expected node data 'force' but was not found in data sample:
 ['entry_id', 'pos', 'atomic_numbers', 'pc_features', 'sizes', 'src_nodes', 'dst_nodes', 'distance_matrix', 'natoms', 'lattice_features', 'targets', 'target_types', 'dataset', 'graph']
```
I think I am getting a better idea but I am still not sure about what the "correct" way is.
Should "force"  be added as an extra entry in the dictionary so that it will be found when transforming to a dgl graph:
  ```      def _copy_node_keys_dgl(self, data: DataDict, graph: DGLGraph) -> None:
            # DGL variant of node data copying
            for key in self.node_keys:
                try:
                    graph.ndata[key] = data[key]
                except KeyError:
                    warn(
                        f"Expected node data '{key}' but was not found in data sample: {list(data.keys())}",
                    )
```
 Right now it's just in ["targets"]["regression"]

**Comment by laserkelvin** [comment]
Thanks for this, I realized I mislead you in my documentation.

So in the `DataDict` being returned, there should be a `targets` and `target_types`. The former should just be flat dictionary with labels, and the latter is a `dict` with regression/classification:

```python
data_dict = {
    ...,
    "targets": {"force": torch.Tensor, "band_gap", 2., "formation_energy": 10., "is_stable": True}
    "target_types": {"regression": ["force"], "classification": ["is_stable"]}
}
```

If you look at [`ForceRegressionTask._get_targets`](https://github.com/IntelLabs/matsciml/blob/ffb68f51522e8e0809b27a9170a08ce6eb8a4543/matsciml/models/base.py#L1433), you'll see how `targets` are being read. If there's any more ambiguity, take a look from one of the other devsets:

```python
>>> from matsciml.datasets import MaterialsProjectDataset

>>> dset = MaterialsProjectDataset.from_devset()
>>> sample = dset.__getitem__(0)
>>> print(sample.keys())
dict_keys(['pos', 'atomic_numbers', 'pc_features', 'sizes', 'src_nodes', 'dst_nodes', 'distance_matrix', 'natoms', 'lattice_features', 'targets', 'target_types', 'dataset'])
```

The warning being raised is when the transform is looking for data to pack into graph structures, and in this case, probably should also look in the `targets` dict as well for tensors. I'm actually open to suggestions as to whether node/edge labels need to be packed into the structure, because @JonathanSchmidt1 is making me realize the way it is now is neither here nor there. I might spin up another issue for this.

laserkelvin:
@JonathanSchmidt1 I've updated the docs with an additional section, "A note on data packing". Hopefully that gives some more clarity, and if you don't have anymore feedback for now I'll merge.

JonathanSchmidt1:
I am very happy with the documentation right now. The only question that is open are per node targets but I assume that will be solved in https://github.com/IntelLabs/matsciml/issues/97 so please merge it.

PR #90, opened 2024-01-10
Running pre-commit on all files

melo-gonzo:
Running pre-commit on all files to get consistent formatting. Updated (stripped down) `.pre-commit-config.yaml` by removing bandit, flake8, and sort-imports.


matsciml/__init__.py
```diff
@@ -1 +1,3 @@
+from __future__ import annotations
+
 __version__ = "1.0.0"
```
laserkelvin, line 3: Bump version to 1.1.0 please!

laserkelvin (changes requested):
Just one change requested, then LGTM :)

matsciml/__init__.py
melo-gonzo: done!

laserkelvin (approved):

PR #103, opened 2024-01-19
OCP Dataset Refactor

melo-gonzo:
This PR refactors the ocp datasets (S2EF, IS2RE) to be `PointCloudDatasets`. This is in preperation of moving to dgl 2.0. The PR includes updated tests and examples, as well as an updated IS2RE devset.


laserkelvin:
Related to #98

examples/model_demos/faenet_pyg.py
```diff
@@ -36,6 +36,11 @@
     "IS2REDataset",
     dset_kwargs={
         "transforms": [
+            PointCloudToGraphTransform(
+                "dgl",
+                cutoff_dist=20.0,
+                node_keys=["pos", "atomic_numbers"],
+            ),
             GraphToGraphTransform("pyg"),
```
laserkelvin, line 44: If we're converting from point cloud to graph, then we should just be able to go straight to PyG right? This `GraphToGraph` is not needed?

examples/model_demos/gala.py
```diff
@@ -23,7 +23,16 @@
 # SchNet uses RBFs, and expects edge features corresponding to atom-atom distances
 dm = MatSciMLDataModule.from_devset(
     "IS2REDataset",
-    dset_kwargs={"transforms": [DistancesTransform()]},
+    dset_kwargs={
+        "transforms": [
+            PointCloudToGraphTransform(
```
laserkelvin, line 28: Wait, why do you need to transform to graphs for GALA?

matsciml/datasets/transforms/representations/tests/test_graph_transform.py
```diff
@@ -105,13 +111,9 @@ def test_dgl_materials_project_fail():
     def test_dgl_ocp_special():
```
laserkelvin, line 111: I don't think this test is needed anymore; I wrote it because we were converting OCP graphs to point clouds via that transform, which we don't need now.

matsciml/datasets/utils.py
```diff
@@ -190,12 +190,12 @@ def point_cloud_featurization(
     torch.Tensor
         Feature tensor, with a shape of [num_src, num_dst, 2 x max_types]
     """
-    eye = torch.eye(max_types)
+    eye = torch.eye(max_types, dtype=int)
```
laserkelvin, line 193: Interesting, where/what kind of error did you get?

I don't have a problem with this, just curious when and where it came out as an issue

laserkelvin (changes requested):

examples/model_demos/gala.py
melo-gonzo: Ha - so looking a bit closer at this, this file has never actually been a Gala demo, it has always had SchNet. I can either make it actually be a gala example, or we can remove the file - Gala is already demonstrated in `./examples/datasets/materials_project/single_task_gala.py`.

laserkelvin: Okay, get rid of it LOL

matsciml/datasets/utils.py
melo-gonzo: not actually seeing any error from this anymore, so i reverted this. cant exactly remember which dataset/pipeline was throwing the error in the first place.

matsciml/datasets/transforms/representations/tests/test_graph_transform.py
melo-gonzo: purged this test

examples/model_demos/faenet_pyg.py
melo-gonzo: good eye - updated this to go straight to pyg.

matsciml/datasets/utils.py
```diff
@@ -195,7 +195,7 @@ def point_cloud_featurization(
     dst_onehot = eye[dst_types][None, :]
     plus, minus = src_onehot + dst_onehot, src_onehot - dst_onehot
     feat_tensor = torch.concat([plus, minus], axis=-1)
-    return feat_tensor
+    return feat_tensor.to(torch.float)
```
laserkelvin, line 198: Pedantic, but code style wise I would prefer `feat_tensor.float()`

laserkelvin (changes requested):
Just made one last comment, then good to go

matsciml/datasets/utils.py
melo-gonzo: done!

laserkelvin (approved):
LGTM

laserkelvin:
Note for future @melo-gonzo for tracking: will need to reprocess IS2RE dataset on Zenodo

PR #108, opened 2024-01-24
Quick Lightning 2.0 Compatibility Update

melo-gonzo:
Adding a default value for dataloader_idx so the MACE example script runs without hiccups.

### DIFF
--- a/matsciml/models/base.py
+++ b/matsciml/models/base.py
@@ -1411,7 +1411,7 @@ class MaceEnergyForceTask(BaseTaskModule):
         return status
 
     def on_validation_batch_start(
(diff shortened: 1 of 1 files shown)


melo-gonzo:
As far as I can tell `dataloader_idx` is only used in `validation/test/predict_batch_start/end`. Is there somewhere you see an inconsistency with this?

laserkelvin (approved):

laserkelvin:
> As far as I can tell `dataloader_idx` is only used in `validation/test/predict_batch_start/end`. Is there somewhere you see an inconsistency with this?

No, was just curious as to how broad the change was within the Lightning API. LGTM, sorry it took a while

PR #109, opened 2024-01-29
Adding support for ColabFit Datasets

EFuem:
This is my initial implementation of a ColabFitDataset. I have a few questions regarding how you'd like things done:

1. Is how I am setting ``target_keys`` ok? We currently have 7 properties supported between all datasets. I didn't see an issue with including all 7 as targets even though datasets will only contain some of these.
2. I didn't see a task in which stresses could be used as a target. Should I remove support for this or is it a plan to support this in the future?

On my end, I still need to streamline the infrastructure for LMDB generation and download for everything we have. I also need to think more deeply about more robust testing/error handling. For example, I believe there are some datasets in which the original creators did not include the same number of potential target properties for all samples, e.g., not including stress for some samples. Being able to warn users about this and other issues would be useful.


melo-gonzo (changes requested):
Thanks for the PR, this looks great! I noted a few small updates. My larger question is if you plan to include an interface to ColabFit in this PR or another? Not sure it makes too much of a difference, just asking for the sake of logistics.

matsciml/datasets/colabfit/dataset.py
```diff
@@ -0,0 +1,109 @@
...
+                data["targets"][key] = data.get(key)
+                data["target_types"]["regression"].append(key)
+        return data
+
+    @property
+    def target_keys(self) -> dict[str, list[str]]:
+        return {"regression": POSSIBLE_KEYS}
+
+    @staticmethod
+    def collate_fn(batch: list[DataDict]) -> BatchDict:
```
melo-gonzo, line 102: Add `BatchDict` to imports.

matsciml/datasets/colabfit/README.md
```diff
@@ -0,0 +1,7 @@
+# ColabFit Database
+
+[ColabFit](https://colabfit.org), a [KIM Initiative project](https://kim-initiative.org) from the University of Minnesota and New York University, seeks to increase the findability, accessibility, interoperability, and reusability of data-driven interatomic potential (DDIP) data by providing convenient access to well-curated and standardized first-principles datasets. Content on the ColabFit Exchange is open source and freely available.
+
+LMDB files can be downloaded via the ColabFit site or through the [colabfit-cli](https://github.com/colabfit/colabfit-cli) tool. The CLI enables querying across all datasets on the ColabFit Exchange and provides an easy interface to download desired datasets in a MatSciML-ready format.
+
+If you utlize the ColabFit Exchange to access data, please cite the original creators of the dataset you use and refer to [this link](https://colabfit.org/about/) for how to cite ColabFit itself.
```
melo-gonzo, line 7: Small typo, utilize.

matsciml/datasets/colabfit/dataset.py
```diff
@@ -0,0 +1,109 @@
...
+        # keep atomic numbers for graph featurization
+        data["atomic_numbers"] = atom_numbers
+        data["pc_features"] = pc_features
+        data["sizes"] = system_size
+        data.update(**node_choices)
+
+        data["targets"] = {}
+        data["target_types"] = {"regression": [], "classification": []}
+        # currently only regression targets present
+        for key in POSSIBLE_KEYS:
```
melo-gonzo, line 91: I think the handling of `target_keys` looks fine. If someone is curating their own ColabFit dataset they should presumably know what keys will be available to them, and if they ask for something not included they'll be met with an error anyway.

melo-gonzo (changes requested):

EFuem:
I have made the small changes.

> Thanks for the PR, this looks great! I noted a few small updates. My larger question is if you plan to include an interface to ColabFit in this PR or another? Not sure it makes too much of a difference, just asking for the sake of logistics.

At the moment there are two ways (or will be) to obtain dataset files-through the website or through a CLI package that I have developed. Obviously the website will always be an option, but whether are not to add the CLI or its components here or just treat it as an external tool (and link to it as is done in the README) is still an open question. There is a third option-support taking in a dataset identifier as an argument to `ColabFitDataset` and then downloading the associated files if not present. However, users would need know what datasets are on ColabFit in the first place (and what the dataset identifiers are) so it may make sense to just default to the first two options.  I am open to suggestions though.

matsciml/datasets/colabfit/dataset.py
```diff
@@ -0,0 +1,109 @@
...
+    "band-gap",
+    "adsorption",
+]
+
+
+@registry.register_dataset("ColabFitDataset")
+class ColabFitDataset(PointCloudDataset):
+    __devset__ = Path(__file__).parents[0].joinpath("devset")
+
+    def index_to_key(self, index: int) -> tuple[int]:
```
laserkelvin, line 27: So this only really needs to be overwritten if you ever intend to just have a single `lmdb` file, but normally this shouldn't be needed (I don't think?). Would you be able to check if this works without overwriting this method?

matsciml/datasets/colabfit/dataset.py
laserkelvin: Am I correct in thinking that `POSSIBLE_KEYS` is based on what *can* be possible from ColabFit, but the point is that the user might not serialize every property?

laserkelvin (approved):

laserkelvin:
Stand alone the PR looks good to me. Just chiming in to the discussion above:

> At the moment there are two ways (or will be) to obtain dataset files-through the website or through a CLI package that I have developed. Obviously the website will always be an option, but whether are not to add the CLI or its components here or just treat it as an external tool (and link to it as is done in the README) is still an open question. There is a third option-support taking in a dataset identifier as an argument to ColabFitDataset and then downloading the associated files if not present. However, users would need know what datasets are on ColabFit in the first place (and what the dataset identifiers are) so it may make sense to just default to the first two options. I am open to suggestions though.

I think we discussed offline last time that IMO it would be good to be able to track the origin of the data within LMDB files: I'm okay to not include the CLI you've written in `matsciml`, but preferably, the CLI will serialize something that includes what datasets, keys, etc. were requested and when, and when the LMDB conversion happens, we dump that information into the `metadata` `dict` as well.

matsciml/datasets/colabfit/dataset.py
EFuem: The devset test works with this removed.

laserkelvin: I would suggest removing it then, just so that the class will work with multiple LMDB files

matsciml/datasets/colabfit/dataset.py
laserkelvin: So I've been debating internally about how to treat this kind of "informational" thing and open to discussions that don't really have to happen inside this PR. Boiling down to _why_ does the user need to know if they're missing keys - as in genuinely, under what circumstances would a user need to know, and what would they be able to do with that information?

This broadly pertains to keys that may or may not be available, like Materials Project. I think the warning message (or even a `KeyError`) should be shown *if* the user specifies it in the `task_keys` specification of a training task, but is missing, since then the user is actually intending to train with it. Otherwise, I don't know if it's needed to have this kind of check within a `Dataset` class?

matsciml/datasets/colabfit/dataset.py
laserkelvin: Done in fbd22d8

melo-gonzo (approved):

PR #113, opened 2024-02-07
Periodic properties transform

laserkelvin:
This PR includes a general transform that will compute edges and distances based on unit cell images. The main use case designed for this is to pipeline with `PointCloudToGraphTransform` to produce graphs that have been wired for periodic boundary conditions, but is designed to be modular (i.e. point clouds could use this too).

A demonstration of its usage is shown in `examples/datasets/periodic_properties_pipeline.py`.

I didn't include it in this PR, but in some ways it supersedes `UnitCellCalculator`. Needs some work to find out how to fully replace it.

The main modifications in `utils.py` are including two functions that A) create Pymatgen `Structure` objects, and B) use `Structure` to compute neighboring atoms while respecting periodic boundary conditions. The new `PeriodicPropertiesTransform` calls these functions, and the results are automatically picked up by the `PointCloudToGraphTransform`, which has been refactored to look for pre-computed edges.

I've tested it on devsets for IS2RE, NOMAD, Materials Project, and LiPS

### DIFF
--- a/examples/datasets/periodic_properties_pipeline.py
+++ b/examples/datasets/periodic_properties_pipeline.py
@@ -1,0 +1,25 @@
+from __future__ import annotations
+
+from matsciml.datasets import IS2REDataset, NomadDataset
+from matsciml.datasets.transforms import (
+    PeriodicPropertiesTransform,
+    PointCloudToGraphTransform,
+)
+
+"""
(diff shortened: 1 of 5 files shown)


melo-gonzo (approved):
LGTM!

PR #119, opened 2024-02-12
Dockerfile update

melo-gonzo:
New dockerfile used for training on GPU.


smiret-intel (approved):
some of these updates could be merged into a smaller ones of lines for future versions

PR #126, opened 2024-02-15
Upgrade matgl to 1.0.0

melo-gonzo:
This PR upgrades the matgl package to 1.0.0. This includes many refactors on their end. In matsciml, the `M3G` datasets are removed, and replaced by a transform, `MGLDataTransform`, which adds the required keys to train m3gnet. Also added in a test for m3gnet specifically.

### DIFF
--- a/matsciml/datasets/transforms/matgl_datasets.py
+++ b/matsciml/datasets/transforms/matgl_datasets.py
@@ -14,6 +14,9 @@ class MGLDataTransform(AbstractDataTransform):
     """
 
     def __call__(self, data: DataDict) -> DataDict:
+        assert (
+            "offsets" in data
+        ), "Offsets missing from data sample! Make sure to include PeriodicPropertiesTransform in the datamodule."
         graph = data["graph"]
         graph.edata["pbc_offset"] = graph.edata["offsets"]
         graph.edata["pbc_offshift"] = torch.matmul(
@@ -23,5 +26,4 @@ class MGLDataTransform(AbstractDataTransform):
         bond_vec, bond_dist = compute_pair_vector_and_distance(graph)
         graph.edata["bond_vec"] = bond_vec
         graph.edata["bond_dist"] = bond_dist
-        data["graph"] = graph
         return data
--- a/matsciml/datasets/transforms/matgl_datasets.py
+++ b/matsciml/datasets/transforms/matgl_datasets.py
@@ -1,0 +1,27 @@
+from __future__ import annotations
+
+from matsciml.common.types import DataDict
+from matsciml.datasets.transforms.base import AbstractDataTransform
+from matgl.graph.compute import compute_pair_vector_and_distance
+import torch
+
+__all__ = ["MGLDataTransform"]
+
+
+class MGLDataTransform(AbstractDataTransform):
+    """
+    Implements a transform to add or swap keys required by matgl models.
+    """
+
+    def __call__(self, data: DataDict) -> DataDict:
+        graph = data["graph"]
+        graph.edata["pbc_offset"] = graph.edata["offsets"]
+        graph.edata["pbc_offshift"] = torch.matmul(
+            graph.edata["pbc_offset"], data["cell"][0]
+        )
+        graph.ndata["node_type"] = graph.ndata["atomic_numbers"].type(torch.int)
+        bond_vec, bond_dist = compute_pair_vector_and_distance(graph)
+        graph.edata["bond_vec"] = bond_vec
+        graph.edata["bond_dist"] = bond_dist
+        data["graph"] = graph
+        return data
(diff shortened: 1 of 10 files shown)


matsciml/models/dgl/tests/test_m3gnet.py
laserkelvin, line 1: Just to be certain, what passes and fails with this test?

matsciml/datasets/transforms/matgl_datasets.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class MGLDataTransform(AbstractDataTransform):
+    """
+    Implements a transform to add or swap keys required by matgl models.
+    """
+
+    def __call__(self, data: DataDict) -> DataDict:
+        graph = data["graph"]
+        graph.edata["pbc_offset"] = graph.edata["offsets"]
```
laserkelvin, line 18: Can you add an assertion here where if this key is missing, it throws a message to remind people to include the transform?

matsciml/datasets/transforms/matgl_datasets.py
```diff
@@ -0,0 +1,27 @@
...
+        graph.edata["pbc_offshift"] = torch.matmul(
```
laserkelvin, line 19: Are these keys the ones that are specifically needed by `matgl`?

matsciml/datasets/transforms/matgl_datasets.py
```diff
@@ -0,0 +1,27 @@
...
+            graph.edata["pbc_offset"], data["cell"][0]
+        )
+        graph.ndata["node_type"] = graph.ndata["atomic_numbers"].type(torch.int)
+        bond_vec, bond_dist = compute_pair_vector_and_distance(graph)
+        graph.edata["bond_vec"] = bond_vec
+        graph.edata["bond_dist"] = bond_dist
+        data["graph"] = graph
```
laserkelvin, line 26: I don't think you need this since we are writing to the graph directly

laserkelvin (changes requested):

matsciml/models/dgl/tests/test_m3gnet.py
melo-gonzo: The OQMD and SyntheticPointGroupDataset datasets will cause this test to fail, due to missing `lattice_params` keys and `atomic_numbers`.
```python
FAILED matsciml/models/dgl/tests/test_m3gnet.py::test_model_forward_nograd[OQMDDataset] - KeyError: 'Data sample is missing lattice parameters. Ensure `lattice_features` or `lattice_params` is available in the data.'
FAILED matsciml/models/dgl/tests/test_m3gnet.py::test_model_forward_nograd[SyntheticPointGroupDataset] - AssertionError: atomic_numbers missing from data sample!
```

matsciml/datasets/transforms/matgl_datasets.py
melo-gonzo: Yes, they are used throughout the codebase, and will be required for TensorNet as well.

matsciml/datasets/transforms/matgl_datasets.py
melo-gonzo: Added assertion in [6ce49cc](https://github.com/IntelLabs/matsciml/pull/126/commits/6ce49cc607ef7be105aeaf8fd6e511ab35a3982b)

matsciml/datasets/transforms/matgl_datasets.py
```diff
@@ -0,0 +1,29 @@
...
+    """
+
+    def __call__(self, data: DataDict) -> DataDict:
+        assert (
+            "offsets" in data
+        ), "Offsets missing from data sample! Make sure to include PeriodicPropertiesTransform in the datamodule."
+        graph = data["graph"]
+        graph.edata["pbc_offset"] = graph.edata["offsets"]
+        graph.edata["pbc_offshift"] = torch.matmul(
+            graph.edata["pbc_offset"], data["cell"][0]
```
laserkelvin, line 23: I'm not 100% sure about the mapping for `pbc_offset` and `pbc_offshift` compared to the results of our transforms: to me, it seems like `pbc_offshift` resembles `offsets`, and it depends on how `pbc_offset` is used, but could be `unit_offsets`?

matsciml/models/dgl/tests/test_m3gnet.py
laserkelvin: Sweet, good to know!

I won't "resolve" this conversation just so that it doesn't get hidden

matsciml/datasets/transforms/matgl_datasets.py
melo-gonzo: After looking a bit closer I think you're right. Their [pbc_offset](https://github.com/materialsvirtuallab/matgl/blob/12ded8c302f9ac9fc23d388bdbd652fb1589cf50/src/matgl/graph/converters.py#L56) is just the images, and the pbc_offshift is our [offsets](https://github.com/melo-gonzo/matsciml/blob/c6dddfe4d99a7abb88fe025c337b735e4cac95c7/matsciml/datasets/utils.py#L719).

Let me know if that sounds right to you.

laserkelvin: I think it seems right. I tried to make my naming convention not as confusing...

Anyhow, make the changes and let's keep this convo open. I'll approve as well so you can merge after you push the change.

laserkelvin (approved):

PR #127, opened 2024-02-16
Periodic graph transform fixes

laserkelvin:
This PR addresses fixes a number of issues pertaining to the periodic property transform workflow:

- An erroneous `in data` condition was interfering with the PyG graph transform under certain circumstances. It was working correctly for like 99% of cases, but in the case that the `all([...])` failed, it would not trigger the branch condition as intended.
- Some structures in Materials Project were failing to produce neighbors, as the structures were incredibly large and nearest atom neighbors were well beyond the 6 Angstrom cutoff. I've added some quasi-helpful exceptions into the mix, as well as added support for a `adaptive_cutoff` flag for the `PeriodicPropertiesTransform` (that gets passed into the subfunction), which will slightly modify the behavior of the neighborhood detection. When this flag is `True`, we will iteratively check for neighbors, incrementing the `cutoff` value by half an Angstrom until we get to a point where every site has a neighbor, or we exceed 30 Angstroms for `cutoff`. A final `neighbors` check will then see catch and throw an error if there are still no neighbors detected (there's something seriously wrong with your structure then).
- I realized that the logging pipeline did not actually capture the `transforms` well at least when using MLFlow as the logger. I modified the `__repr__` method of `AbstractTransform` to return transforms in a dictionary representation, which hopefully will give enough information to recreate the transform at a later time. It would be good to make sure this doesn't impact other loggers.

### DIFF
--- a/matsciml/datasets/transforms/base.py
+++ b/matsciml/datasets/transforms/base.py
@@ -34,3 +34,6 @@ class AbstractDataTransform:
             Transformed item from an abstract dataset
         """
         raise NotImplementedError
+
+    def __repr__(self) -> str:
+        return f"{self.__class__.__name__}: {self.__dict__}"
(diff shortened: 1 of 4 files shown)


melo-gonzo (approved):
Looks good - these changes will be helpful.

PR #128, opened 2024-02-16
M3GNet Model Refactor

melo-gonzo:
With the update to matgl to include [intermediate outputs](https://github.com/materialsvirtuallab/matgl/pull/212) in m3gnet, the required embeddings for the matsciml Embeddings object are easy to grab, and the full forward pass does not need to be copied over.

### DIFF
--- a/matsciml/models/dgl/m3gnet.py
+++ b/matsciml/models/dgl/m3gnet.py
@@ -29,6 +29,7 @@ class M3GNet(AbstractDGLModel):
         self.elemenet_types = element_types
         self.all_embeddings = return_all_layer_output
         self.model = matgl_m3gnet(element_types, *args, **kwargs)
+        self.atomic_embedding = self.model.embedding
 
     def _forward(
         self,
--- a/matsciml/models/dgl/m3gnet.py
+++ b/matsciml/models/dgl/m3gnet.py
@@ -43,5 +43,6 @@ class M3GNet(AbstractDGLModel):
         outputs = self.model(
             graph, return_all_layer_output=self.all_embeddings, **kwargs
         )
-        # gc_3 is essentially the last graph layer before the readout
-        return Embeddings(outputs["readout"], outputs["gc_3"]["node_feat"])
+        # gc_{self.model.n_blocks} is essentially the last graph layer before the readout
+        last_layer = f"gc_{self.model.n_blocks}"
+        return Embeddings(outputs["readout"], outputs[last_layer]["node_feat"])
--- a/matsciml/models/dgl/m3gnet.py
+++ b/matsciml/models/dgl/m3gnet.py
@@ -22,9 +22,12 @@ and to construct the Embedding's output object.
 
 @registry.register_model(M3GNet)
 class M3GNet(AbstractDGLModel):
-    def __init__(self, element_types, *args, **kwargs):
+    def __init__(
+        self, element_types: list[str], return_all_layer_output: bool, *args, **kwargs
+    ):
         super().__init__(atom_embedding_dim=len(element_types))
         self.elemenet_types = element_types
+        self.all_embeddings = return_all_layer_output
         self.model = matgl_m3gnet(element_types, *args, **kwargs)
 
     def _forward(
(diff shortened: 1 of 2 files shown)


matsciml/models/dgl/m3gnet.py
```diff
@@ -32,49 +32,10 @@ def forward(
...
-        vec = torch.hstack([node_vec, state_feat]) if self.include_states else node_vec  # type: ignore
-        output = self.final_layer(vec)
-        if self.task_type == "classification":
-            output = self.sigmoid(output)
-    else:
-        g.ndata["atomic_properties"] = self.final_layer(g)
-        output = dgl.readout_nodes(g, "atomic_properties", op="sum")
-    return Embeddings(vec, node_vec)
+    graph = batch["graph"]
+    outputs = self.m3gnet_forward(graph, state_attr, l_g, return_all_layer_output=True)
```
laserkelvin, line 36: ❤️

laserkelvin (changes requested):
So honestly if we're going to refactor, we could do the end-all-be-all and use inheritance like the other model definitions:

```python
from matgl.models import M3GNet as matgl_m3gnet

@registry.register_model(...)
class M3GNet(AbstractDGLModel):
     ...
     self.model = matgl_m3gnet(*args, **kwargs)

     def forward(self, *args, **kwargs):
         return self.model(*args, **kwargs)
```

Not sure about the feasibility, but does mean that it would have consistent behavior to the other models (i.e. if we modified the base classes)

What do you think?

matsciml/models/dgl/m3gnet.py
```diff
@@ -5,38 +5,36 @@
...
-
-
-M3GNet.m3gnet_forward = M3GNet.forward
-M3GNet.forward = forward
+@registry.register_model(M3GNet)
+class M3GNet(AbstractDGLModel):
+    def __init__(self, element_types, *args, **kwargs):
+        super().__init__(atom_embedding_dim=len(element_types))
+        self.elemenet_types = element_types
+        self.model = matgl_m3gnet(element_types, *args, **kwargs)
```
laserkelvin, line 28: If `matgl` has its own embedding table, it might be good to overwrite one so we don't have duplicates. Probably keep the `M3GNet` one

```python
# I don't know what they're called so please check!
self.atomic_embedding = self.model.embedding
```

laserkelvin (changes requested):

melo-gonzo:
> What do you think?

This makes a lot of sense and looks a lot better. I've updated the model to use matsciml style inheritance.

matsciml/models/dgl/m3gnet.py
```diff
@@ -1,81 +1,47 @@
...
-    else:
-        g.ndata["atomic_properties"] = self.final_layer(g)
-        output = dgl.readout_nodes(g, "atomic_properties", op="sum")
-    return Embeddings(vec, node_vec)
-
-
-M3GNet.m3gnet_forward = M3GNet.forward
-M3GNet.forward = forward
+        # gc_3 is essentially the last graph layer before the readout
+        return Embeddings(outputs["readout"], outputs["gc_3"]["node_feat"])
```
laserkelvin, line 47: I swear this is the last comment so if you address this feel free to merge, but this numbering doesn't change with hyperparameters right?

laserkelvin (approved):
Left the last question, merge when done

matsciml/models/dgl/m3gnet.py
```diff
@@ -1,81 +1,48 @@
...
+from matsciml.models.base import AbstractDGLModel
+
+"""
+M3GNet is integrated from matgl: https://github.com/materialsvirtuallab/matgl
+The forward pass needs to be overwritten to accommodate matsciml's expected inputs,
+and to construct the Embedding's output object.
+"""
+
+
+@registry.register_model(M3GNet)
```
laserkelvin, line 23: Crap, this should be a string!

laserkelvin (changes requested):

matsciml/models/dgl/m3gnet.py
melo-gonzo: stringing: [c9dedac](https://github.com/IntelLabs/matsciml/pull/128/commits/c9dedac7f4377d38e0499dea32da3fcf03f1d1cc)

laserkelvin (approved):

PR #129, opened 2024-02-16
TensorNet Integration

melo-gonzo:
This PR adds TensorNet from matgl. The helpful `return_all_layer_output` is not yet available for this model so the forward pass needed to be copied in and modified to use matsciml `Embeddings` as an output. Additionally, the `TensorEmbedding` needed to be updated to accept a different number of elements. The model is implemented using the inheritance structure expected in matsciml, and will make refactoring the model easier when/if `return_all_layer_output` is added to the model on the matgl side. A minimal example and model test are also included. Note that [these same two](https://github.com/IntelLabs/matsciml/pull/126#discussion_r1491664770) test fail as expected.

Note that there is an alternative way to implement TensorNet (similar to the first pass of M3GNet), where embeddings do not need to be updated, but the inheritance scheme is not utilized: [alternative implementation](https://github.com/IntelLabs/matsciml/blob/d0903a95707e08ee3344dc8ca504061f9614e179/matsciml/models/dgl/tensornet.py).


matsciml/models/dgl/tests/test_tensornet.py
```diff
@@ -0,0 +1,61 @@
...
+        registry.__entries__["datasets"].keys(),
+    ),
+)
+
+
+@pytest.mark.parametrize(
+    "dset_class_name",
+    filtered_list,
+)
+def test_model_forward_nograd(dset_class_name: str, model_fixture: TensorNet):
```
laserkelvin, line 38: Same things pass as before?

matsciml/models/dgl/tensornet.py
```diff
@@ -0,0 +1,90 @@
...
+            state_attr: State attrs for a batch of graphs.
+            **kwargs: For future flexibility. Not used at the moment.
+
+        Returns:
+            output: output: Output property for a batch of graphs
+        """
+        # Obtain graph, with distances and relative position vectors
+        g.edata["pbc_offshift"] = g.edata["offsets"]
+        g.ndata["node_type"] = g.ndata["atomic_numbers"].type(torch.int)
+        bond_vec, bond_dist = compute_pair_vector_and_distance(g)
```
laserkelvin, line 52: 😿

matsciml/models/dgl/tensornet.py
```diff
@@ -0,0 +1,90 @@
...
+            g : DGLGraph for a batch of graphs.
+            state_attr: State attrs for a batch of graphs.
+            **kwargs: For future flexibility. Not used at the moment.
+
+        Returns:
+            output: output: Output property for a batch of graphs
+        """
+        # Obtain graph, with distances and relative position vectors
+        g.edata["pbc_offshift"] = g.edata["offsets"]
+        g.ndata["node_type"] = g.ndata["atomic_numbers"].type(torch.int)
```
laserkelvin, line 51: Can you do like `.long()` instead?

matsciml/models/dgl/tests/test_tensornet.py
melo-gonzo: Exactly!

matsciml/models/dgl/tensornet.py
melo-gonzo: We could always remove this stuff, and use the `MGLDataTransform`, but I think it makes sense to just have the forward pass match up with matgl side for now.

matsciml/models/dgl/tensornet.py
melo-gonzo: I sure [can](https://github.com/IntelLabs/matsciml/pull/129/commits/4729731efe20413febbae77c3eaa2828b79ad60a)

matsciml/models/dgl/tensornet.py
```diff
@@ -0,0 +1,90 @@
...
+from matgl.graph.compute import (
+    compute_pair_vector_and_distance,
+)
+from matgl.models import TensorNet as MGLTensorNet
+from matgl.utils.maths import decompose_tensor, tensor_norm
+
+from matsciml.common.types import Embeddings
+from matsciml.common.registry import registry
+from matsciml.models.base import AbstractDGLModel
+
```
laserkelvin, line 15: Last thing, I think you need a `__all__` here and add the model name

laserkelvin (approved):

matsciml/models/dgl/tensornet.py
melo-gonzo: [Done](https://github.com/IntelLabs/matsciml/pull/129/commits/c0a4519dd9e6e5edead1ae600e1e865a8e8cb9d1)

PR #130, opened 2024-02-16
Code carbon integration

laserkelvin:
This PR introduces a `CodeCarbonCallback` to the stack of PyTorch Lightning callbacks in `matsciml`, which integrates `codecarbon` usage throughout training, testing, and inference `pl.Trainer` workflows.

### DIFF
--- a/examples/callbacks/code-carbon.py
+++ b/examples/callbacks/code-carbon.py
@@ -27,7 +27,7 @@ dm = MatSciMLDataModule.from_devset(
     "IS2REDataset",
     dset_kwargs={
         "transforms": [
-            PeriodicPropertiesTransform(6.0),
+            PeriodicPropertiesTransform(6.0, adaptive_cutoff=True),
             PointCloudToGraphTransform(
                 "pyg",
                 node_keys=["pos", "atomic_numbers"],
--- a/examples/callbacks/code-carbon.py
+++ b/examples/callbacks/code-carbon.py
@@ -1,0 +1,44 @@
+from __future__ import annotations
+
+import pytorch_lightning as pl
+
+from matsciml.datasets.transforms import (
+    PointCloudToGraphTransform,
+    PeriodicPropertiesTransform,
+)
+from matsciml.lightning.callbacks import CodeCarbonCallback
+from matsciml.lightning.data_utils import MatSciMLDataModule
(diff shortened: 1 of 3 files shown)


examples/callbacks/code-carbon.py
melo-gonzo, line 1: I get an error with the `PeriodicProperteisTransform` in this example - we should include the `adaptive_cutoff` flag or bump up the radius.

matsciml/lightning/callbacks.py
melo-gonzo, line 1: Can you add the [`on_exception`](https://github.com/Lightning-AI/pytorch-lightning/blob/682d7ef6b4f8ab412e523f161d95d3e88fbe58cf/src/lightning/pytorch/callbacks/callback.py#L228) method as well to capture the `emissions.pt` data if something kills training, or it's hit with a KeyboardInterrupt?

matsciml/lightning/callbacks.py
```diff
@@ -537,3 +538,140 @@ def write_on_epoch_end(
...
+
+            Parameters
+            ----------
+            trainer
+                Instance of ``pl.Trainer``, which is used to access the
+                logger instances as well.
+            """
+            log_dir = self._temp_output_dir
+            # if nothing was specified
+            if not log_dir:
```
melo-gonzo, line 627: Can these two `if`'s be combo'd into one line?

melo-gonzo (changes requested):
Looks good overall! Will be good data to have. Left a few comments.

matsciml/lightning/callbacks.py
laserkelvin: Addressed in b32e680

matsciml/lightning/callbacks.py
laserkelvin: Addessed in b20ff2b

examples/callbacks/code-carbon.py
laserkelvin: Fixed in 8c8dc4e

laserkelvin:
Thanks for all the great feedback! The `on_exception` suggestion actually got me thinking to rewrite it in a much better way too!

melo-gonzo:
Thanks for the clean-ups, approved!

melo-gonzo (approved):

PR #132, opened 2024-02-19
Add alexandria dataset to matsciml toolkit https://github.com/IntelLabs/matsciml/discussions/107

JonathanSchmidt1:
This pull request adds the AlexandriaRequest class to download the Alexandria database to an lmdb dataset and a AlexandriaDataset class to use it for ML within the matsciml toolkit.
The AlexandriaDataset class mostly follows  the Materials project example.
I also added most of the same tests.
At the moment there are not test for  the AlexandriaRequest class.


examples/model_demos/mpnn_alexandria_dgl.py
melo-gonzo, line 1: Thanks for including an example! Since this is a new dataset, the example may be better placed in `examples/datasets/alexandria/single_task_devset.py`.

matsciml/datasets/alexandria/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+The [alexandria database](https://alexandria.icams.rub.de/) is maintained by Miguel Marques at RUB.
+The database comprises  ~4.5 million three-dimensional relaxed crystal structures that span
+the periodic table, in addition to over 100,000 two-dimensional and 10,000 one-dimensional crystal structures.
+Further ~400k of the three-dimensional crystal structures are also available at PBEsol geometries and with all properties
+calculated with the SCAN functional.
+Each structures has an associated total energy, band gap, magnetization, magnetic moments on each atom,
+distance to the convex hull per atom, formation energy per atom and density of states at the fermi level.
+A fixed training, validation and test split as well as a link to a FAIR repository will be added in the future
+with a further publication.
```
laserkelvin, line 11: Do you mind adding a list of target keys here and give a quick summary of each property - e.g. what units they use, and anything else that might be relevant?

matsciml/datasets/alexandria/README.md
melo-gonzo, line 1: Can you also add a basic script snippet to show how to download the full dataset? Similar to [Carolina Materials Database](https://github.com/melo-gonzo/matsciml/tree/main/matsciml/datasets/carolina_db#:~:text=To%20download%20the%20full%20dataset%2C%20the%20CMDRequest%20module%20from%20matsciml/datasets/carolina_db/carolina_api.py%20may%20be%20used%20as%20follows%3A) or [OQMD](https://github.com/melo-gonzo/matsciml/tree/main/matsciml/datasets/oqmd#:~:text=A%20simple%20python%20script%20may%20be%20used%20to%20query%2C%20save%20and%20process%20the%20data%20into%20an%20lmdb%20file%3A). Additionally, it would be idea to have the default full dataset download options included into the [`./scripts/download_datasets.py`](https://github.com/melo-gonzo/matsciml/blob/main/scripts/download_datasets.py) script as well.

matsciml/datasets/alexandria/dataset.py
```diff
@@ -0,0 +1,237 @@
...
+        system_size = len(coords)
+        return_dict["pos"] = coords
+        chosen_nodes = self.choose_dst_nodes(system_size, self.full_pairwise)
+        src_nodes, dst_nodes = chosen_nodes["src_nodes"], chosen_nodes["dst_nodes"]
+        atom_numbers = torch.LongTensor(structure.atomic_numbers)
+        # uses one-hot encoding featurization
+        pc_features = point_cloud_featurization(
+            atom_numbers[src_nodes],
+            atom_numbers[dst_nodes],
+            110,
```
laserkelvin, line 81: Is there a reason you went with 110 here?

matsciml/datasets/alexandria/api.py
```diff
@@ -0,0 +1,172 @@
...
+from pathlib import Path
+from typing import Any, Dict, List, Optional, Union
+
+import numpy as np
+import requests
+from joblib import Parallel, delayed
+from tqdm import tqdm
+
+from matsciml.datasets import utils
+from matsciml.datasets.alexandria import AlexandriaDataset
```
melo-gonzo, line 15: There's some unused imports in here that can be removed.

matsciml/datasets/alexandria/dataset.py
```diff
@@ -0,0 +1,237 @@
...
+class AlexandriaDataset(PointCloudDataset):
+    __devset__ = Path(__file__).parents[0].joinpath("devset")
+
+    @property
+    def target_keys(self) -> dict[str, list[str]]:
+        # This returns the standardized dictionary of task_type/key mapping.
+        keys = getattr(self, "_target_keys", None)
+        if not keys:
+            # grab a sample from the data to set the keys
+            _ = self.__getitem__(0)
```
laserkelvin, line 33: This is more of a style comment than anything: normally `_` is set when you don't care about the variable but you just need to unpack it, but you do end up using it in the case below so I'd prefer it you named it like `sample` or something

matsciml/datasets/alexandria/dataset.py
melo-gonzo, line 81: @laserkelvin do you know if having different `max_types` for the point_cloud_featurization` will cause problems in a multi-data scenario? I can test this, but it's not something we've done so far.

matsciml/datasets/alexandria/dataset.py
```diff
@@ -0,0 +1,237 @@
...
+        self._parse_structure(data, return_dict)
+        self._parse_symmetry(data, return_dict)
+        return_dict["force"] = self._standardize_values(data["force"])
+        return_dict["magmoms"] = self._standardize_values(data["magmoms"])
+
+        targets = {
+            key: self._standardize_values(data["targets"]["regression"][key])
+            for key in data["targets"].get("regression", {}).keys()
+        }
+        targets = targets | {
```
laserkelvin, line 173: Does this statement work on our lowest version of Python @melo-gonzo?

It might work with `from __future__ import annotations` and that's definitely true for type annotations but I wasn't sure if it works as an operator. Can you try running a snippet like this in a container you use?

matsciml/datasets/alexandria/dataset.py
```diff
@@ -0,0 +1,237 @@
...
+        """
+        data: dict[str, Any] = super().data_from_key(lmdb_index, subindex)
+        return_dict = {"entry_id": data["entry_id"]}
+        # parse out relevant structure/lattice data
+        self._parse_structure(data, return_dict)
+        self._parse_symmetry(data, return_dict)
+        return_dict["force"] = self._standardize_values(data["force"])
+        return_dict["magmoms"] = self._standardize_values(data["magmoms"])
+
+        targets = {
```
laserkelvin, line 169: For the sake of readibility I might actually suggest calling the two `dict`s as `regression_targets` and `classification_targets`. You can still do 

```python
targets = regression_targets | classification_targets
```

But this way IMO the intention is more clear even if it's more verbose

matsciml/datasets/alexandria/dataset.py
```diff
@@ -0,0 +1,237 @@
...
+from matsciml.datasets.base import PointCloudDataset
+from matsciml.datasets.utils import (
+    concatenate_keys,
+    element_types,
+    point_cloud_featurization,
+)
+
+
+@registry.register_dataset("AlexandriaDataset")
+class AlexandriaDataset(PointCloudDataset):
```
laserkelvin, line 24: More a comment for discussion with @melo-gonzo but the fact that @JonathanSchmidt1 had to copy over a bunch of functions here means we could do better on the abstraction: what do you think?

matsciml/datasets/alexandria/tests/test_alexandria.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+def test_dataset_collate():
+    dset = AlexandriaDataset(AlexandriaDataset.__devset__)
+    data = [dset.__getitem__(index) for index in range(10)]
+    batch = dset.collate_fn(data)
+    # check the nuclear coordinates and numbers match what is expected
+    assert batch["pos"].shape[-1] == 3
+    assert batch["pos"].ndim == 2
+    assert len(batch["atomic_numbers"]) == 10
```
laserkelvin, line 23: Is this right? One atom per batch sample.

Just making sure!

matsciml/datasets/alexandria/tests/test_alexandria.py
```diff
@@ -0,0 +1,103 @@
...
+            for key in ["pos", "pc_features", "sizes", "src_nodes", "dst_nodes"]
+        ],
+    )
+    # for a pairwise point cloud sizes should be equal
+    feats = sample["pc_features"]
+    assert feats.size(0) >= feats.size(1)
+    assert sample["pos"].ndim == 2
+
+
+def test_graph_transform():
```
laserkelvin, line 95: Would it be a big ask for you to add a PyG transform test as well?

laserkelvin (changes requested):
Thanks @JonathanSchmidt1 for the great PR!!

For the most part things look good and functional to me; I did leave a few style remarks, and wanted to see if you could add a PyG test as well. If you don't normally use that pipeline, don't worry about it and we can add it later.

melo-gonzo (changes requested):
Thanks for the PR! It looks pretty good overall, I've left a few initial comments. In addition to what was noted, can you add some details to the [DATASETS.md](https://github.com/melo-gonzo/matsciml/tree/main/matsciml/datasets#documenting-new-datasets:~:text=Documenting%20new%20datasets) as well? It's a bit redundant but for consistency it will be nice to have some details in there as well.

laserkelvin:
LOL the timing on our reviews

matsciml/datasets/alexandria/dataset.py
laserkelvin: Yeah it would cause problems if it's used naively, but the caveat is if we're using point clouds, this featurization might actually be done by the model `_read_batch` step anyway in order to have it backprop through to the atomic numbers (which is why IIRC this is [done here](https://github.com/IntelLabs/matsciml/blob/7afb766a093bf1fb3c18d733707726472088070d/matsciml/models/base.py#L335).) 

If anything it might mean we don't need to do this step in the data loading, which we can refactor out later.

examples/model_demos/mpnn_alexandria_dgl.py
JonathanSchmidt1: thanks I added examples similar to the materials project and moved them

matsciml/datasets/alexandria/dataset.py
JonathanSchmidt1: changed _ to sample

JonathanSchmidt1:
Thank you for both of the reviews. I will go through them today and tomorrow.
I noticed one issue that the way periodic boundary conditions are dealt with was updated over the last weeks when my graphtransform test failed.
How is the output of a periodic graph supposed to look after the transformation with [PeriodicPropertiesTransform(20.), PointCloudToGraphTransform('pyg', cutoff_dist=20.0)] now?
i.e. Are the edges supposed to show the edges including boundary conditions already? Or is this calculated somewhere else later based on the shifts and offsets (if yes where?).

btw maybe it's an issue for the materials project dataset also but for my dataset the check for precomputed src and dst nodes lead to there only being self loops in the end 
```
g.edge_index
tensor([[0, 1, 2],
        [0, 1, 2]])

```
instead of
```
tensor([[1, 2, 2],
        [0, 0, 1]])
```
with the check commented out

JonathanSchmidt1:
ok the first question solved itself. Now I am just wondering about the edge case if you do not use the PeriodicPropertiesTransform
but the materials project or Alexandria dataset you already have src/dst nodes set from the PointCloudDataset choose_dst_nodes method. However these do not seem to be the correct edges to me e.g. with a cutoff of 40 I would expect connections between the different atoms:
```
>>> dset.__getitem__(0)['graph'].edge_index
tensor([[0, 1, 2],
        [0, 1, 2]])
>>> dset.__getitem__(0)['graph'].pos
tensor([[ 0.0000,  0.0000,  0.0000],
        [22.0659,  4.5494,  2.7011],
        [ 6.3406,  1.3073,  0.7761]])

```
but maybe I am missunderstanding the edge indices

matsciml/datasets/alexandria/dataset.py
JonathanSchmidt1: The parse_structure/parse_symmetry functions from the materials project class seems quite general. Maybe it makes sense to put them in some utils file instead and import them in the respective datasets.

matsciml/datasets/alexandria/tests/test_alexandria.py
JonathanSchmidt1: I added two pyg transform tests https://github.com/IntelLabs/matsciml/pull/132/commits/b1ce3af847843cfd7dca4b4f592ab6849ed0c247

matsciml/datasets/alexandria/dataset.py
JonathanSchmidt1: Fixed it to 100

matsciml/datasets/alexandria/README.md
JonathanSchmidt1: I added a download script example in the dataset examples and added options for the various datasets to the script.

laserkelvin:
> but in the materials project or Alexandria dataset you already have src/dst nodes set from the PointCloudDataset choose_dst_nodes method. However these do not seem to be the correct edges to me e.g. with a cutoff of 40 I would expect connections between the different atoms (this is related to the second point in the previous question):

To be clear (I don't know if this is a misunderstanding or not so forgive me) the function `choose_dst_nodes` is probably poorly named, because point clouds don't have edges. That function is used to construct featurizations for point clouds but are not related to *graph edge construction*, so the cut-off radii do not come into play at all. That's why I think `edge_index` don't make sense

If you want to understand the graph edge construction, there are basically two implemented paths now: the `cutoff_radius` value is applied either by an explicit dense matrix distance calculation (if there is no `cell` attribute in the data sample, i.e. assume it's basically like a molecular graph) or if there is `cell`, we rely on `pymatgen` to compute the connectivity.

matsciml/datasets/alexandria/dataset.py
melo-gonzo: Can confirm that this doesn't work in python <= 3.8. It would be good to switch up these statements for compatibility.

matsciml/datasets/alexandria/dataset.py
melo-gonzo: Agreed, a lot of this can be refactored into some sort of utils file - it would make all the datasets much cleaner. I can draft up an issue for this.

matsciml/datasets/alexandria/dataset.py
laserkelvin: In fairness, Python 3.8 is end of life this year so maybe we should just move to Python 3.9+ as supported versions. That would be a separate PR however.

For this specific PR my recommendation is to rewrite it without using the syntatic sugar.

JonathanSchmidt1: changed it to
```
classification_targets = {
            key: self._standardize_values(data["targets"]["classification"][key])
            for key in data["targets"].get("classification", {}).keys()
        }
targets = targets.update(classification_targets)``

**Comment by JonathanSchmidt1** [comment]
@laserkelvin One question concerning the test that I did not completely understand is still open. The rest should be ready for rereview.

**Comment by laserkelvin** [comment]
I'm out of office for the rest of the week and I'll review it once I get back.

File: matsciml/datasets/alexandria/api.py [hunk 10]
```diff
@@ -0,0 +1,180 @@
...
+                raise ValueError(
+                    f"Request failed with status code {response.status_code}",
+                )
+
+        data = download(self.urls[index])
+        computed_entry_dict = [
+            self.get_data_dict(entry)
+            for entry in tqdm(data["entries"], desc=f"Processing file {index}")
+        ]
+        if self.devset:
```
laserkelvin, line 150: Would it make sense to move this up so you don't run all of the requests for devsets? That way you don't use too much server load if you don't need to.

matsciml/datasets/alexandria/dataset.py
```diff
@@ -0,0 +1,236 @@
...
+            Output dictionary that contains the training sample. Mutates
+            in place.
+
+        Raises
+        ------
+        ValueError:
+            If `Structure` is not found in the data; currently the workflow
+            intends for you to have the data structure in place.
+        """
+        structure: None | Structure = Structure.from_dict(data.get("structure", None))
```
laserkelvin, line 65: If you are unpacking a `pymatgen.core.Structure`, could you also add the `Lattice` object to the data dict?

This should get reused by the `PeriodicPropertiesTransform` and make it consistent across datasets.

i.e.

```python
cell = torch.from_numpy(structure.lattice.matrix).float()
return_dict["cell"] = cell
```

matsciml/datasets/alexandria/tests/test_alexandria.py
laserkelvin: Ahh sorry, in my head I thought the tensor was padded and concatenated. My mistake!

matsciml/datasets/alexandria/api.py
JonathanSchmidt1: I can move it up one step  so we run self.get_data_dict only for the 100 entries. Right now we only have the database split up into the 100k structure jsons so one has to download at least one of them (and it also will only  be one of them for the devset so that should not be an issue for the server).
```
     def devset(cls, lmdb_target_dir: str) -> AlexandriaRequest:
        return cls([0], lmdb_target_dir, dataset="3D", devset=True)
```

matsciml/datasets/alexandria/__init__.py
```diff
@@ -0,0 +1,12 @@
+from __future__ import annotations
+
+from pathlib import Path
+from matsciml.datasets.alexandria.api import AlexandriaRequest
+from matsciml.datasets.alexandria.dataset import AlexandriaDataset
+
+alexandria_devset = Path(__file__).parents[0].joinpath("devset")
```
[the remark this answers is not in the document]
JonathanSchmidt1: :+1:

matsciml/datasets/alexandria/dataset.py
JonathanSchmidt1: :+1:   added the cell to the return dict

examples/datasets/alexandria/single_task_devset.py
```diff
@@ -0,0 +1,31 @@
...
+# configure a simple model for testing
+model = GraphConvModel(100, 128, encoder_only=True)
+task = ScalarRegressionTask(model, task_keys=["band_gap_ind"])
+
+# configure alexandria devset
+dm = MatSciMLDataModule.from_devset(
+    "AlexandriaDataset",
+    dset_kwargs={
+        "transforms": [
+            PeriodicPropertiesTransform(20.0),
```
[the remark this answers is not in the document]
JonathanSchmidt1: I reduced it to 10

examples/datasets/alexandria/single_task_egnn.py
```diff
@@ -0,0 +1,82 @@
...
+    lr=1e-3,
+    task_keys=["band_gap_ind"],
+)
+
+dm = MatSciMLDataModule(
+    dataset="AlexandriaDataset",
+    train_path="../../../matsciml/datasets/alexandria/devset",
+    dset_kwargs={
+        "transforms": [
+            PeriodicPropertiesTransform(20.0),
```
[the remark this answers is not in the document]
JonathanSchmidt1: I reduced it to 10

examples/datasets/alexandria/single_task_base.py
```diff
@@ -0,0 +1,46 @@
...
+        "activation": SiLU,
+        "lazy": False,
+        "input_dim": 128,
+    },
+    lr=1e-3,
+    task_keys=["band_gap_ind"],
+)
+
+
+dm = MatSciMLDataModule(
```
melo-gonzo, line 32: Can you update the examples to use `MatSciMLDataModule.from_devset`? Using this means also taking out the `val_split` as well.

examples/datasets/alexandria/single_task_egnn.py
```diff
@@ -0,0 +1,79 @@
...
+        "hidden_dim": 128,
+        "activation": SiLU,
+        "lazy": False,
+        "input_dim": 128,
+    },
+    lr=1e-3,
+    task_keys=["band_gap_ind"],
+)
+
+dm = MatSciMLDataModule(
```
melo-gonzo, line 58: Update to use `.from_devset`

examples/datasets/alexandria/single_task_gala.py
```diff
@@ -0,0 +1,58 @@
...
+        "activation": SiLU,
+        "lazy": False,
+        "input_dim": 128,
+    },
+    lr=1e-4,
+    task_keys=["band_gap_ind"],
+)
+
+
+dm = MatSciMLDataModule(
```
melo-gonzo, line 49: Update to use `.from_devset`

melo-gonzo (commented):
A few more minor updates from my end, and then will be good to approve!

laserkelvin (approved):
I just highlighted the unused import flagged by `pre-commit`, but other than that I'm giving my approval for merging.

melo-gonzo (approved):

laserkelvin:
I think we're good to go - I looked into the failing tests and nothing is relevant to this PR.

Thank you @JonathanSchmidt1 for all your hard work on this PR!

PR #134, opened 2024-02-21
Adding FA Transform Back In

melo-gonzo:
Frame averaging transform was removed in [this](https://github.com/IntelLabs/matsciml/commit/34d2677c3c51357f1a35f1574dd36e22758f7142) commit by ruff. There doesn't seem to be a great solution to prevent ruff from removing unused imports, and also for allowing * imports. I added `# noqa: F403` to get around this for now.

### DIFF
--- a/matsciml/datasets/transforms/frame_averaging.py
+++ b/matsciml/datasets/transforms/frame_averaging.py
@@ -17,12 +17,11 @@ from matsciml.datasets.transforms.base import AbstractDataTransform
 
 __all__ = ["FrameAveraging"]
 
-if package_registry["dgl"]:
-    pass
-
 if package_registry["pyg"]:
     import torch_geometric
     from torch_geometric.transforms import LinearTransformation
+else:
+    raise Exception("torch_geometric is required for Frame Averaging transform.")
 
 
 def compute_frames(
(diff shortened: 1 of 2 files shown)


matsciml/datasets/transforms/frame_averaging.py
```diff
@@ -15,12 +15,14 @@
...
 
-if package_registry["dgl"]:
-    import dgl
+__all__ = ["FrameAveraging"]
 
 if package_registry["pyg"]:
     import torch_geometric
     from torch_geometric.transforms import LinearTransformation
+else:
+    raise Exception("torch_geometric is required for Frame Averaging transform.")
```
laserkelvin, line 24: You want to probably use `ModuleNotFoundError` instead.

As a side note, it's more that it's just implemented for PyG right now and not DGL right? I would suggest "Frame averaging transform is only currently implemented for PyG."

laserkelvin (changes requested):

matsciml/datasets/transforms/frame_averaging.py
melo-gonzo: updated in [d212fdf](https://github.com/IntelLabs/matsciml/pull/134/commits/d212fdf7170fb63b81d02ffdfb93dcc302d6778d)

laserkelvin (approved):

laserkelvin:
Apparently 2 reviewers are needed for this review, so I added @smiret-intel

smiret-intel (approved):

PR #135, opened 2024-02-22
Update MaterialsProjectDataset for mp-traj and gnome compatibility

melo-gonzo:
mp-traj and gnome add in some new targets, a few are: force and stress. This small update will let the existing MaterialsProjectDataset be usable across all three datasets (mp, mp-traj, gnome). Note that the `magmom` target is giving problems at the moment and is disregarded for now.

### DIFF
--- a/matsciml/datasets/materials_project/dataset.py
+++ b/matsciml/datasets/materials_project/dataset.py
@@ -242,7 +242,13 @@ class MaterialsProjectDataset(PointCloudDataset):
         self._parse_symmetry(data, return_dict)
         # assume every other key are targets
         not_targets = set(
-            ["structure", "symmetry", "fields_not_requested", "formula_pretty"]
+            [
+                "structure",
+                "symmetry",
+                "fields_not_requested",
+                "formula_pretty",
+                "magmom",
+            ]
             + data["fields_not_requested"],
         )
         # target_keys = getattr(self, "_target_keys", None)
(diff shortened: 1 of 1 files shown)


laserkelvin:
I'm a little confused looking at the code and your description and just wanted to make sure something has changed as intended: I only see force and stress being added as default collate keys, but is there more involved to add to actually add the labels (i.e. into the LMDB files)?

melo-gonzo:
the new datasets, materials project trajectory and gnome, already have the labels included in the lmdb files. specifically the target keys they have are: 
```json
"mp-traj": [
    "uncorrected_total_energy",
    "corrected_total_energy",
    "energy_per_atom",
    "ef_per_atom",
    "e_per_atom_relaxed",
    "ef_per_atom_relaxed",
    "force",
    "stress",
    "magmom",
    "bandgap",
],
"gnome": ["corrected_total_energy", "force"],
```

the changes here ensure that these two new datasets can be run with the existing MaterialsProjectDataset. hopefully that helps clarify.

laserkelvin:
I think because this assumes new LMDB files, can we add some kind of check that if the keys aren't there, we throw a `KeyError` and tell the user to make sure they have the newest LMDB files from the Zenodo?

This affects us (when we're using it) and others as well

laserkelvin (changes requested):
I think because this assumes new LMDB files, can we add some kind of check that if the keys aren't there, we throw a KeyError and tell the user to make sure they have the newest LMDB files from the Zenodo?

This affects us (when we're using it) and others as well

melo-gonzo:
I'm getting a bit confused now as well. We can still use the code as it is in the PR for any of the 3 datasets: Materials Project, Materials Project Trajectory, and Gnome. Are you wanting a check for the trajectory and gnome keys specifically? The keys between each three datasets are not all the same, similar to how the ColabFit datasets will not necessarily have all the same keys.

laserkelvin:
Ahh okay, my mistake. Never mind I understand now!

laserkelvin (approved):

smiret-intel (approved):

PR #141, opened 2024-03-08
Updating oqmd dataset, devset, and example

melo-gonzo:
After generating new dataset splits, OQMD changed its `unit_cell` key to just `cell` for consistency with other datasets. The `lattice_params` are also added into the dataset. The `oqmd/single_task_devset.py` is also updated to show `PeriodicPropertiesTransform` used with oqmd data.


matsciml/datasets/oqmd/dataset.py
```diff
@@ -89,7 +90,15 @@ def data_from_key(
         return_dict["atomic_numbers"] = atom_numbers
         return_dict["pc_features"] = pc_features
         return_dict["sizes"] = system_size
-        return_dict["unit_cell"] = unit_cell
+        return_dict["cell"] = cell
```
laserkelvin, line 93: Is this line duplicated?

laserkelvin (changes requested):

matsciml/datasets/oqmd/dataset.py
melo-gonzo: Fixed [37f46da](https://github.com/IntelLabs/matsciml/pull/141/commits/37f46da223676689d07bac3010d1d3429e6d2699)

PR #142, opened 2024-03-08
Nomad Dataset and Devset Update

melo-gonzo:
Updating devset and dataset to include new regression keys" `"reference_energy", "reference_structure", "relative_energy"`. Updated nomad example script to use one of the new keys.


matsciml/datasets/nomad/devset/devset_ids.yml
laserkelvin, line 1: Why was this file deleted?

laserkelvin (changes requested):

matsciml/datasets/nomad/devset/devset_ids.yml
melo-gonzo: Those id's no longer reflect what is in the devset (first 200 samples of train lmdb). Also, none of the other datasets have this file so it may be confusing in the future.

laserkelvin (approved):
LGTM

PR #144, opened 2024-03-12
Adding `pre-commit` workflow to Github actions

laserkelvin:
This PR adds a workflow to run `pre-commit` on the repository upon pull request to `main`.

### DIFF
--- a/.github/workflows/pre-commit-action.yml
+++ b/.github/workflows/pre-commit-action.yml
@@ -1,0 +1,18 @@
+name: Run pre-commit workflow
+on:
+  pull_request:
+    branches:
+      - main
+jobs:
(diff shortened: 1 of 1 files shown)


laserkelvin:
So a bunch of stuff fail, mainly pertain to obsolete modules and whatnot that should get fixed but I think are out of scope for this PR.

IMO if there aren't any other behavior-related stuff I say we merge it to make sure formatting/linting is actually run.

laserkelvin:
Disregard the last comment, per @melo-gonzo's suggestion the action will only run `pre-commit` on files changed since last PR which is a lot faster and efficient!

melo-gonzo (approved):
Thanks!

PR #145, opened 2024-03-12
Tests refactor

laserkelvin:
This PR heavily refactors the majority of tests in `matsciml`, primarily to use semantic `pytest.mark` to determine which tests to include and which to skip.

The main purpose of this is to prepare for CI running of tests, where we want to:

1. Skip really slow tests, including heavy models and remote API requests
2. Selectively run tests based on settings, e.g. distributed or not

Not every test passes at this time, but I've fixed the ones that I've deemed to be "minimal". All of the tests were run with the following command from the root directory:

```console
pytest -v -m "not remote_request and not lmdb and not slow"
```


melo-gonzo (approved):
A super useful update, thanks!

PR #146, opened 2024-03-12
Selective PyTest workflow on Github actions

laserkelvin:
This PR creates workflows for selectively running PyTest based on contents of pull requests.

The idea is that certain tests are only based on changes in submodules, i.e. we only run dataset related tests if datasets are changed, etc.

Not entirely sure about performance right now, but it uses `micromamba` to create and cache environments based on the `conda.yml` in the root folder.


melo-gonzo (approved):
Looks fantastic

PR #147, opened 2024-03-12
Fix for unused atomic embedding table in M3Gnet

### LINKED ISSUE #137: [Bug]: unused parameters in  m3gnet
### Expected behavior

I would expect the model not to have any unused parameters during the backward process.

### Actual behavior

Running ddp without strategy find_unused_parameters crashes.
My theory is that 
```
class M3GNet(AbstractDGLModel):
    def __init__(
        self, element_types: list[str], return_all_layer_output: bool, *args, **kwargs
    ):
        super().__init__(atom_embedding_dim=len(element_types))
        self.elemenet_types = element_types
        self.all_embeddings = return_all_layer_output
        self.model = matgl_m3gnet(element_types, *args, **kwargs)
        self.atomic_embedding = self.model.embedding
```
self.atomic_embedding should be self.atom_embedding as that is the name in the AbstractDglTask.
As the m3gnet embedding block is more than just node embeddings it should probably just be:
`        self.atom_embedding = self.model.embedding.layer_node_embedding`
This fixed it for me (I also removed the final m3gnet layer so maybe that also has to be done).

Run m3gnet with ddp.

### Specifications

the latest matsciml version with dgl 1.1.3

laserkelvin:
Thanks to @JonathanSchmidt1 for opening the issue!

This PR closes #137, which had an unintended bug where the M3GNet embedding table is not correctly mapped to the correct attribute name expected by `AbstractDGLModel`. Consequently, you end up with unused parameters that are picked up when DDP is used.

Per Jonathan's suggestion, this PR just changes the attribute mapping so that the atom table is correctly referenced and no unused parameters are present.

### DIFF
--- a/matsciml/models/dgl/m3gnet.py
+++ b/matsciml/models/dgl/m3gnet.py
@@ -27,7 +27,7 @@ class M3GNet(AbstractDGLModel):
         self.elemenet_types = element_types
         self.all_embeddings = return_all_layer_output
         self.model = matgl_m3gnet(element_types, *args, **kwargs)
-        self.atomic_embedding = self.model.embedding
+        self.atom_embedding = self.model.embedding.layer_node_embedding
 
     def _forward(
         self,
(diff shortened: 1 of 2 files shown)


melo-gonzo (approved):
This is a good catch - thanks for the fix!

PR #148, opened 2024-03-12
Add minimal end-to-end workflow

laserkelvin:
This PR adds (for now) a simple EGNN end-to-end PyTest and accompanying Github actions workflow definition.

The premise for this is to have at least one example of an end-to-end run, which hopefully preserves functionality regardless of the changes being submitted are. 

I'm open to adding more tests, but this was the simplest one I could think of and is usually my "go to".


melo-gonzo (approved):
LG 🫡

PR #153, opened 2024-03-14
Tensornet Embedding Output Update

melo-gonzo:
This PR updates the embedding object required by matsciml. Previously, and incompatible node embedding was used causing problems with `GradFreeForceRegression` due to tensor size mismatches. Can @smiret-intel and @laserkelvin look to see if this update makes sense for the node embedding?

### DIFF
--- a/matsciml/models/dgl/tests/test_tensornet.py
+++ b/matsciml/models/dgl/tests/test_tensornet.py
@@ -85,7 +85,7 @@ def test_force_regression(model_fixture, devset_fixture):
     task = ForceRegressionTask(
         model_fixture, output_kwargs={"lazy": False, "input_dim": 64, "hidden_dim": 64}
     )
-    trainer = pl.Trainer(max_steps=5, logger=False, enable_checkpointing=False)
+    trainer = pl.Trainer(fast_dev_run=10)
     trainer.fit(task, datamodule=devset_fixture)
     # make sure losses are tracked
     for key in ["energy", "force"]:
@@ -97,7 +97,7 @@ def test_gradfree_force_regression(model_fixture, devset_fixture):
         model_fixture,
         output_kwargs={"lazy": False, "input_dim": 64, "hidden_dim": 64},
     )
-    trainer = pl.Trainer(max_steps=5, logger=False, enable_checkpointing=False)
+    trainer = pl.Trainer(fast_dev_run=10)
     trainer.fit(task, datamodule=devset_fixture)
     # make sure losses are tracked
     assert "train_force" in trainer.logged_metrics
--- a/matsciml/models/dgl/tests/test_tensornet.py
+++ b/matsciml/models/dgl/tests/test_tensornet.py
@@ -1,4 +1,5 @@
 import pytest
+import pytorch_lightning as pl
 
 from matsciml.datasets.transforms import (
     PeriodicPropertiesTransform,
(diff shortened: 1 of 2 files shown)


laserkelvin:
It's hard to tell based on their implementation of `tensor_norm` which dimensions are actually being reduced, and what shape you end up with. 

Also, is the model being used with `intensive=True`? At least looking at the code, it seems like `node_vec` is produced with `readout`, which to me, should be the system embeddings and not the node-level embeddings (as you might expect) which I think are actually `x`...

I looked at the tests, and it seems that at least the forward pass remains functional. Can you write a unit test that uses `TensorNet` with `S2EF` devset and `GradFreeForceRegression` to make sure it works?

melo-gonzo:
I've added tests for `ForceRegressionTask` and `GradFreeForceRegressionTask`, which both pass.

The pipeline is using `is_intensive`. For some more context, the embedding size of `X` is `[batch size, hidden embedding size, 3, 3]`, and `tensor_norm(X)` is `[batch size, hidden embedding size]`. The dimensions of `x` are `[batch size, 3*hidden embedding size]`. 

The concatenation of decomposed  `X` is what is throwing me off most: `scalars, skew_metrices, traceless_tensors = decompose_tensor(X)`. As far as compatible embedding dimensions goes, what this PR introduces works, I am just unsure about the correctness of the embedding used.

matsciml/models/dgl/tests/test_tensornet.py
```diff
@@ -59,3 +79,25 @@ def test_model_forward_nograd(dset_class_name: str, model_fixture: TensorNet):
         assert ~torch.isnan(z).all()  # check there are no NaNs
         assert torch.isfinite(z).all()
         assert torch.all(torch.abs(z) <= 1000)  # ensure reasonable values
+
+
+def test_force_regression(model_fixture, devset_fixture):
+    task = ForceRegressionTask(
+        model_fixture, output_kwargs={"lazy": False, "input_dim": 64, "hidden_dim": 64}
+    )
+    trainer = pl.Trainer(max_steps=5, logger=False, enable_checkpointing=False)
```
laserkelvin, line 88: Can you do `fast_dev_run` instead, or do you need to in order for the key checking assertion to work?

laserkelvin (commented):
Thanks for adding the tests!

So I don't think `tensor_norm(X)` should be shape `[batch_size, hidden_embedding_size]`.

The `Embedding` structure we have should take a graph-level and node-level embedding pair, right, and so I'd expect:

- `vec` is shape `[batch_size, hidden_embedding_size]`
- `node_vec` (or `tensor_norm(X)`) should be `[num_nodes, hidden_embedding_size]`

melo-gonzo:
Ok sorry, so that was my bad - `tensor_norm(X)` is in fact `[num_nodes, hidden_embedding_size]`.`node_vec` is [batch_size, hidden_embedding_size], which is unexpected and why it hasn't been working with `GradFreeForceRegression`. And `vec` is as expected, `[batch_size, hidden_embedding_size]`.

matsciml/models/dgl/tests/test_tensornet.py
melo-gonzo: updated here:[0ef2cf4](https://github.com/IntelLabs/matsciml/pull/153/commits/0ef2cf41153a8be6e4df4e7d46974cc0ac4e14a3)

laserkelvin (approved):
tensornet tests pass, and the failing ones aren't relevant. LGTM

PR #154, opened 2024-03-15
Enable Intel XPU usage and examples

laserkelvin:
This PR adds PyTorch Lightning classes that enable the usage of Intel GPUs, formerly codenamed Ponte Vecchio, with the `matsciml` pipeline.

An example script is provided, which provides the minimum working configuration on a single device.


laserkelvin:
The tests that failed in the lightning test suite aren't relevant to this PR, should be good to go.

smiret-intel (approved):

PR #167, opened 2024-04-03
Callback for Sharpness Aware Minimization 

Vaibhav525:
Implements sharpness-aware minimization as PyTorch lightening callback

### DIFF
--- a/matsciml/lightning/callbacks.py
+++ b/matsciml/lightning/callbacks.py
@@ -8,15 +8,16 @@ from datetime import datetime
 from logging import DEBUG, getLogger
 from pathlib import Path
 from time import time
-from typing import Any, Callable
+from typing import Any, Callable , Dict ,Iterator, Optional
 
 import numpy as np
 import pytorch_lightning as pl
 import torch
+from pytorch_lightning import LightningModule, Trainer
 from pytorch_lightning.callbacks import BasePredictionWriter, Callback
 from pytorch_lightning.utilities.rank_zero import rank_zero_only
 from torch import distributed as dist
-from torch import nn
+from torch import nn 
 from torch.optim import Optimizer
 
 from matsciml.common.packages import package_registry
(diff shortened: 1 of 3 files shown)


laserkelvin:
@Vaibhav525 the CI checks fail. Some of the Lightning ones are unrelated, but the pre-commit workflow should be green. The issues that are coming up are to do with code style, so please install pre-commit hooks and address the issues? I will try and review the implementation in parallel.

In addition, please add some unit tests to make sure that the functionality of SAM. I'm not sure if this will work on every task based on how it's implemented, specifically on how it assumes a single optimizer. I know that the original paper assumes a single optimizer/task, but I would be very curious to see how this affects property prediction.

The other broader comment that I haven't really thought too much about yet (I will do so when I review it) is whether or not this is better as a callback, or a task. I would appreciate if @melo-gonzo gave his opinion on this as well.

matsciml/lightning/callbacks.py
```diff
@@ -683,3 +684,113 @@ def on_exception(
...
+        self,
+        trainer: Trainer,
+        pl_module: LightningModule,
+        batch: Any,
+        batch_idx: int,
+    ) -> None:
+        self.batch = batch
+        self.batch_idx = batch_idx
+
+    @torch.no_grad()
```
laserkelvin, line 748: If the reason for decorating with `no_grad` is because of the `first_step` call, then I would suggest making it consistent with the `torch.enable_grad` context usage.

matsciml/lightning/callbacks.py
```diff
@@ -683,3 +684,113 @@ def on_exception(
...
+        self,
+        trainer: Trainer,
+        pl_module: LightningModule,
+        optimizer: Optimizer,
+    ) -> None:
+        org_weights = self._first_step(optimizer)
+        with torch.enable_grad():
+            step_output = pl_module.training_step(self.batch, self.batch_idx)
+            loss = self._get_loss(step_output)
+            if loss is not None:
```
laserkelvin, line 759: What circumstances would `loss` be `None`?

matsciml/lightning/callbacks.py
```diff
@@ -683,3 +684,113 @@ def on_exception(
...
+        return param_norms.norm()
+
+    def _first_step(self, optimizer: Optimizer) -> Dict[torch.Tensor, torch.Tensor]:
+        scale = self.rho / (self._grad_norm(optimizer) + 1e-5)
+        org_weights: Dict[torch.Tensor, torch.Tensor] = {}
+        for p in self._get_params(optimizer):
+            if p.grad is None:
+                continue
+            org_weights[p] = p.detach().clone()
+            e_w = (torch.pow(p, 2) if self.adaptive else 1.0) * p.grad * scale.to(p)
```
laserkelvin, line 785: Why do you need `scale.to(p)`? I would have thought that all tensors should be on the same device.

matsciml/lightning/callbacks.py
```diff
@@ -683,3 +684,113 @@ def on_exception(
...
+                (self._norm_weights(p) * p.grad).norm()
+                for p in self._get_params(optimizer)
+                if isinstance(p.grad, torch.Tensor)
+            ]
+        )
+        return param_norms.norm()
+
+    def _first_step(self, optimizer: Optimizer) -> Dict[torch.Tensor, torch.Tensor]:
+        scale = self.rho / (self._grad_norm(optimizer) + 1e-5)
+        org_weights: Dict[torch.Tensor, torch.Tensor] = {}
```
laserkelvin, line 780: This is a bit weird to me, since you are indexing tensors in a dictionary with the tensor itself. I don't know how stable this is, and if it works for now you can keep it but please add some comments for the code.

melo-gonzo:
To Kelvin's point about implementing as a task vs callback - seeing that `pl_module.training_step` is called, this will run the full forward pass logic to only get the same information that you would otherwise get from the task module. To me this would mean implementing this as a standalone task may make more sense, however the current callback implementation makes it easy to pop in and out of the existing tasks. I'm not sure how easily you could write this as a standalone task and then apply it to each of the existing task we already have, so I'm ok with keeping this as a callback, and reassessing if there seems to be any problems from doing through the forward pass twice. Also, I'm unsure how this would work with our multi-optimizer setups and would need a closer look if we plan to do experiments in that setting. 

Can @laserkelvin link to existing tests for the last two checkmarks for some examples of how to add those to the repo? I know there are some examples out there somewhere already.

matsciml/lightning/callbacks.py
```diff
@@ -683,3 +684,113 @@ def on_exception(
...
+        trainer: Trainer,
+        pl_module: LightningModule,
+        optimizer: Optimizer,
+    ) -> None:
+        org_weights = self._first_step(optimizer)
+        with torch.enable_grad():
+            step_output = pl_module.training_step(self.batch, self.batch_idx)
+            loss = self._get_loss(step_output)
+            if loss is not None:
+                trainer.strategy.backward(
```
laserkelvin, line 760: Can you confirm whether or not this is actually what the SAM paper intended?

image

The reason I ask is that my cursory glance at the SAM paper suggests hessians are involved even though they aren't fully materialized. Their pseudocode matches your implementation, but I wanted to make sure that we aren't dropping a `create_graph=True` in the `backward` call when we need it, or vice versa.

laserkelvin (changes requested):

laserkelvin:
> To Kelvin's point about implementing as a task vs callback - seeing that `pl_module.training_step` is called, this will run the full forward pass logic to only get the same information that you would otherwise get from the task module. To me this would mean implementing this as a standalone task may make more sense, however the current callback implementation makes it easy to pop in and out of the existing tasks.\

Actually so reading through the paper and @Vaibhav525's implementation I think this as a callback is fine. You _do_ need to run the forward pass again to get essentially the second derivative of the loss w.r.t. inputs, and is used to compute some approximation to the SAM gradient.

> Also, I'm unsure how this would work with our multi-optimizer setups and would need a closer look if we plan to do experiments in that setting.

Yeah this is my concern as well, so as you say we should ponder about how/what we want to train.

> Can @laserkelvin link to existing tests for the last two checkmarks for some examples of how to add those to the repo? I know there are some examples out there somewhere already.

[This test on EGNN](https://github.com/IntelLabs/matsciml/blob/main/matsciml/tests/test_pipeline.py) is something I would have in mind. Ideally we run it on a matrix of models and datasets, but for now, I think showing that the full double forward/backward passes run in a single model and dataset setting would be suffice. I'll let @Vaibhav525 decide between `SchNet` which he used for the example, or `EGNN` for consistency (as my personal choice). That said, I think the more interesting cases would be like `FAENet`/MACE.

My recommendation would be then to add three unit tests with EGNN, FAENet, and MACE on whatever dataset. For MACE, use the parameters in [this example](https://github.com/IntelLabs/matsciml/blob/main/examples/model_demos/mace_wrapper.py) and same for [FAENet](https://github.com/IntelLabs/matsciml/blob/main/examples/model_demos/faenet_pyg.py).

matsciml/lightning/callbacks.py
Vaibhav525: I agree. It is not required.

matsciml/lightning/callbacks.py
Vaibhav525: It is not likely to be None, but it depends on the rho parameter, if it is very large it can become nan or inf. But I guess we can remove the if condition

matsciml/lightning/callbacks.py
Vaibhav525: Yes it is done for the first-step, can you suggest how to change it?

Vaibhav525 (commented):
I have addressed the comments. Kindly have a look

laserkelvin:
@Vaibhav525 a few pre-commit checks are still failing - did you install and run them?

matsciml/lightning/callbacks.py
laserkelvin: If it can be `NaN` or `inf` then I think it's more appropriate to check for that, instead of using `None`

matsciml/lightning/callbacks.py
laserkelvin: Just wrap the first step call like you do the rest of the model:

```python
with torch.no_grad():
    org_weights = self._first_step(optimizer)

with torch.enable_grad():
   ...

with torch.no_grad():
    self._second_step(...)
```

laserkelvin (changes requested):

Vaibhav525:
Pre-commit check is now successful. I will add the unit tests as discussed soon.

laserkelvin:
Thank you! Please ping us again when it's ready

Vaibhav525:
@laserkelvin I have added tests for SAM in the `test_sam.py` file. As anticipated by you, the `ForceRegressionTask` test is not successful.

matsciml/lightning/tests/test_sam.py
```diff
@@ -0,0 +1,303 @@
...
+    Tests force regression with SAM using PLEGNNBackbone.
+    """
+    devset = MatSciMLDataModule.from_devset(
+        "S2EFDataset",
+        dset_kwargs={
+            "transforms": [
+                PointCloudToGraphTransform(
+                    "dgl",
+                    cutoff_dist=20.0,
+                    node_keys=["pos", "atomic_numbers"],
```
laserkelvin, line 167: Possibly this is failing because it's missing the `force` key

laserkelvin (changes requested):

matsciml/lightning/tests/test_sam.py
Vaibhav525: The error is encountered in `on_before_optimizer_step` of SAM, without SAM all tests are passing.

laserkelvin:
I discussed with @smiret-intel and we agreed that it is probably easiest to try and merge this in now since almost everything is working, and we know that force regression does not. Given that you have tests in place now it's easier for me/us to try and get it working.

laserkelvin (approved):

melo-gonzo (approved):

PR #187, opened 2024-04-18
SAM Callback DDP and Multi Data Fixes.

melo-gonzo:
Two bugs are addressed in this PR:

1. In DDP, the pytorch lightning `trainer` api is different, so the original `trainer.model.optimizers` will not work properly. This pr switches the logic to use the task module to get the optimizer names instead, which is consistent across training pipelines.

2. In multi-data training, each dataset does not necessarily have the same amount of samples. When there is sample imbalance, the `_compute_loss()` function only returns losses for datasets that are still being processed. The `optimizer_names` still contain mappings for datasets with no more samples to process, which was causing problems when gathering losses and computing the "global loss". This pr adds a few checks to `extract_optimizer_specific_loss` to fix these issues.

### DIFF
--- a/matsciml/lightning/callbacks.py
+++ b/matsciml/lightning/callbacks.py
@@ -783,6 +783,7 @@ class SAM(Callback):
             org_weights = self._first_step(optimizer)
         with torch.enable_grad():
             loss = task._compute_losses(self.batch)
+            # this is for the multitask case where there is more than on optimizer
             if len(task.optimizers()) > 1:
                 loss = self.extract_optimizer_specific_loss(task, optimizer, loss)
             loss = self._get_loss(loss)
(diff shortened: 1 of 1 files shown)


matsciml/lightning/callbacks.py
```diff
@@ -779,11 +783,12 @@ def on_before_optimizer_step(
             org_weights = self._first_step(optimizer)
         with torch.enable_grad():
             loss = task._compute_losses(self.batch)
-            if len(trainer.optimizers) > 1:
-                loss = self.extract_optimizer_specific_loss(trainer, optimizer, loss)
+            if len(task.optimizers()) > 1:
```
laserkelvin, line 786: Could you just throw a comment in here just to say this is the multitask case?

laserkelvin (approved):
Just one comment - merge after you've added

PR #191, opened 2024-04-19
Noisy node positions pretraining task

laserkelvin:
This PR adds a pretraining task to `matsciml` as described in [Zaidi _et al._](https://openreview.net/pdf?id=tYIMtogyee), where an encoder/output head is used to predict the noise injected into the input atom positions.

- An example script is given, under `examples/tasks/pretraining` that demonstrates its usage on EGNN+Alexandria dataset
- Implements a `NoisyPositions` data transform inside a `transforms.pretraining` module, which is intended to house other future pre-training transformations
    - Included unit tests that span datasets and representations; seems to be all green across supported datasets, point cloud, graphs
- Implements a `NodeDenoisingTask`, which handles the training.
    - Included a single unit test using DGL EGNN and Materials Project

### DIFF
--- a/examples/tasks/pretraining/denoising_nodes.py
+++ b/examples/tasks/pretraining/denoising_nodes.py
@@ -18,6 +18,12 @@ pretraining task, as described in:
 Pre-training via denoising for molecular property prediction
 
 by Zaidi _et al._, ICLR 2023; https://openreview.net/pdf?id=tYIMtogyee
+
+The permutation of transforms is not fully invariant, based on how
+it is currently implemented. This configuration is the recommended one,
+where positions are noised _after_ generating the periodic properties;
+this is to ensure that the periodic offsets are generated based on
+the noise-free positions.
 """
 
 # construct IS2RE relaxed energy regression with PyG implementation of E(n)-GNN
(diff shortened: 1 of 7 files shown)


laserkelvin:
@melo-gonzo seems like Lightning is inconsistent with what it returns with `pl_module.optimizers`: #187 I think assumes it was a list which is true for multitask, but for single tasks it just returns the optimizer.

Do you want to submit a PR to fix this, or I can as well

melo-gonzo:
> @melo-gonzo seems like Lightning is inconsistent with what it returns with `pl_module.optimizers`: #187 I think assumes it was a list which is true for multitask, but for single tasks it just returns the optimizer.
> 
> Do you want to submit a PR to fix this, or I can as well

Feel free to submit the fix

melo-gonzo (commented):
This looks great overall. My only concern with the implementation is in using the `NoisyPositions` transform before the `PeriodicPropertiesTransform`. When used before, the noisy positions do not get transformed to fractional coordinates which I believe are expected by some architectures like M3GNet. The fractional coordinates are also used to compute properties used by M3GNet (and perhaps others, cannot remember off the top of my head) such as `unit_offsets`. 

Simply switching around the order of the transforms would eliminate this concern.

laserkelvin:
> Simply switching around the order of the transforms would eliminate this concern.

I've changed it in the unit tests in 09c3316, and for the example in 92da14e

PR #192, opened 2024-04-19
Correcting SAM check for multitask case

laserkelvin:
Previous check would fail for the single task case, as Lightning will not return a list but the optimizer directly.

The check now looks to see if the actual optimizer is returned, and if not, will then run the multitask check.

Issue originally showed up in #191 CI checks.


melo-gonzo (approved):
Thanks!

laserkelvin:
I think #193 makes this redundant right?

If so please close this without merging!

laserkelvin:
Okay yeah, this is redundant now so I'm going to close it.

PR #194, opened 2024-04-19
ForceRegressionTask Update for Frame Averaging Compatibility

melo-gonzo:
The current `ForceRegressionTask` assumes that the `pos` variable will be uses for computing gradients. In Frame Averaging, a transformed version of the position coordinates are used and these updated coordinates should also be used for computing the forces. This PR introduces some logic to handle this for the `stochastic` and `all` Frame Averaging cases (others may work, but not tested).


examples/model_demos/faenet_pyg.py
```diff
@@ -18,22 +16,22 @@
 """
 
 # construct IS2RE relaxed energy regression with PyG implementation of FAENet
```
laserkelvin, line 18: Update comment since it's S2EF

examples/model_demos/faenet_pyg.py
```diff
@@ -48,43 +46,5 @@
...
-            ),
-            FrameAveraging(frame_averaging="3D", fa_method="stochastic"),
-        ],
-    },
-)
-
-# run a quick training loop
-trainer = pl.Trainer(fast_dev_run=10)
-trainer.fit(task, datamodule=dm)
+trainer = pl.Trainer(enable_checkpointing=False, logger=False, devices=1)
```
laserkelvin, line 49: Is there a reason you're not using `fast_dev_run`?

matsciml/models/base.py
```diff
@@ -1564,53 +1568,71 @@ def forward(
...
-            -1
-            * torch.autograd.grad(
-                energy,
-                pos,
-                grad_outputs=torch.ones_like(energy),
-                create_graph=True,
-            )[0]
-        )
+
+        def energy_and_force(pos, system_embedding):
```
laserkelvin, line 1593: I think it's clear from the variable names, but can you add type annotations anyway?

laserkelvin (changes requested):

laserkelvin:
Model pytests failing are unrelated to this PR, and the `pre-commit` hooks that fail are just stylistic ones addressed in #191

examples/model_demos/faenet_pyg.py
melo-gonzo: updated in [502337c](https://github.com/IntelLabs/matsciml/pull/194/commits/502337cacf13a7e2a83aa7297ebfca7a6a3a1cbe)

examples/model_demos/faenet_pyg.py
melo-gonzo: updated in [502337c](https://github.com/IntelLabs/matsciml/pull/194/commits/502337cacf13a7e2a83aa7297ebfca7a6a3a1cbe)

laserkelvin (approved):

matsciml/models/base.py
melo-gonzo: updated in [502337c](https://github.com/IntelLabs/matsciml/pull/194/commits/502337cacf13a7e2a83aa7297ebfca7a6a3a1cbe)

Issue #166, opened 2024-03-25
[Feature request]: PyG installation instructions (esp. for XPUs)


chaitjo:
### Feature/behavior summary

I'm trying to get PyG to install and work well with Intel XPUs, and was hoping to use this repository as reference. At present, I see that PyG is never installed by default, and nor are any instructions for setting it up with XPUs available.

### Request attributes

- [ ] Would this be a refactor of existing code?
- [ ] Does this proposal require new package dependencies?
- [ ] Would this change break backwards compatibility?
- [ ] Does this proposal include a new model?
- [ ] Does this proposal include a new dataset?
- [ ] Does this proposal include a new task/workflow?

### Solution description

Unknown.

### Additional notes

At present, working with a different repository (https://github.com/a-r-j/ProteinWorkshop), I've been trying to integrate your code for the XPU as a new accelerator in PyTorch Lightning: https://github.com/IntelLabs/matsciml/blob/main/matsciml/lightning/xpu.py.

So far, I'm able to get my trainer to identify the XPU as a device, but it seems like some torch_cluster operations are not compatible with tensor stored on XPUs. I would like to perform torch_cluster operations such as knn graph creation on XPU tensors so that I can do data processing in a batched manner or on-the-fly, as opposed to on the CPU. 

Here is a minimal example which fails:
```
import torch
import intel_extension_for_pytorch as ipex
from torch_geometric.nn import knn_graph

device = torch.device('xpu:0' if torch.xpu.is_available() else 'cpu')

x = torch.tensor([[-1.0, -1.0], [-1.0, 1.0], [1.0, -1.0], [1.0, 1.0]]).to(device)
batch = torch.tensor([0, 0, 0, 0]).to(device)
edge_index = knn_graph(x, k=2, batch=batch, loop=False)
```
The resulting error is `RuntimeError: x.device().is_cpu() INTERNAL ASSERT FAILED at "csrc/cpu/knn_cpu.cpp":12, please report a bug to PyTorch. x must be CPU tensor`.

And here's a longer trace from the ProteinWorkshop codebase, which probably won't make any sense to MatSciML maintainers.
```
File "/home/ckj24/rds/hpc-work/envs/proteinworkshop/lib/python3.10/site-packages/torch_geometric/nn/pool/__init__.py", line 171, in knn_graph
    return torch_cluster.knn_graph(x, k, batch, loop, flow, cosine,
[...]

PR #198, opened 2024-04-23
Revised XPU instructions and features

laserkelvin:
This PR partially addresses #166, and ensures a degree of OOB experience with Intel XPUs.

The main entry point are scripts in the examples/devices folder that demonstrate DDP functionality on XPUs on managed clusters, using both Slurm as a launcher as well as Intel MPI. To enable this, the provided conda.yml and updated readme gives guidance on how to go from environment creation to looking at the examples.

- Separated environment specification to `conda.yml` and `conda-xpu.yml` for CPU-only and XPU builds (because they're different for reasons). The former is used for the CI pipeline.
- Attempted to make the Lightning abstractions as seamless as possible, but still not perfect because of some hard to pin down issues between XPU and Lightning DDP.


laserkelvin:
@smiret-intel @melo-gonzo this is now ready for review

melo-gonzo (approved):

PR #204, opened 2024-05-03
Node embedding projections for `ForceRegressionTask`

laserkelvin:
This PR refactors `ForceRegressionTask` to regress energies by reducing node contributions, as opposed to system-level embedding regression.

- Energy output head sees node embeddings now, rather than graph embeddings
- Define `readout` functions that depend on whether a graph structure is passed, and what framework is used. The reduction can be configured with the _task_ hyperparameter `embedding_reduction_type` 
- `process_embeddings` will pack the node energy contributions into the output dictionary.


laserkelvin:
The force task tests pass

melo-gonzo (approved):
LGTM! Just out of curiosity, but what would the use case be for using a readout other than 'sum' for a graph level energy?

laserkelvin:
You can also do a mean reduction, or if you want to be super fancy, an attention weighted one (like in graph attention). Sum is more physically meaningful than a mean reduction IMO

PR #205, opened 2024-05-03
Frame Averaging Energy Reduction Bugfix

melo-gonzo:
Currently, when frame averaging is uses, the predicted energy values are stored in a list because each frame will have an associated prediction. When [reduced](https://github.com/melo-gonzo/matsciml/blob/07f440923ff41e7177dd1b3ad4a891d165e56893/matsciml/models/base.py#L1637), this would collapse all of the predictions into a single scalar value, instead of a scalar value per sample in a batch. This PR addresses this issue by concatenating all predicted energy outputs prior to the reduction.

This PR also adds tests for FAENet for this specific bug as well as general forward pass checks for the applicable datasets. 

This PR adds a check to keys in a batch, and includes `natoms` if not present, inside of the `read_batch` of FAENet itself.


laserkelvin:
This doesn't seem to have the stuff from #204 - can you rebase?

matsciml/models/pyg/faenet/faenet.py
```diff
@@ -251,15 +251,20 @@ def read_batch(self, batch: BatchDict) -> DataDict:
...
-
         data = {"graph": batch.get("graph")}
         graph = batch.get("graph")
         for key in ["edge_feats", "graph_feats"]:
             data[key] = getattr(graph, key, None)
         pos: torch.Tensor = getattr(graph, "pos")
         data["pos"] = pos
         data["graph"].cell = batch["cell"]
-        data["graph"].natoms = batch["natoms"].squeeze(-1).to(torch.int32)
+        if "natoms" not in batch:
```
laserkelvin, line 261: Actually, I feel like if you have access to the graph, I would rather you read the number of atoms from there instead; I feel like it would be a lot more readable that way than from `ptr`

matsciml/models/base.py
```diff
@@ -1667,8 +1667,9 @@ def energy_and_force(
                     force[frame_idx].view(-1, 1, 3).bmm(repeat_rot.transpose(1, 2))
                 )
                 all_forces.append(rotated_forces)
-            # combine all the force data into a single tensor
+            # combine all the force and energy data into a single tensor
```
laserkelvin, line 1670: Could you annotate the expected shapes from both energy and force?

laserkelvin (changes requested):

matsciml/models/pyg/faenet/faenet.py
melo-gonzo: This is number of atoms per sample - I don't think there is an easy to grab graph attribute for that unless I'm missing something.

matsciml/models/base.py
melo-gonzo: added a note here [2a61886](https://github.com/IntelLabs/matsciml/pull/205/commits/2a6188612e0458a7813ff2cf89bfbe3743f12cf1)

matsciml/models/pyg/faenet/faenet.py
laserkelvin: For DGL you can do like [`graph.batch_num_nodes()`](https://docs.dgl.ai/en/1.1.x/generated/dgl.DGLGraph.batch_num_nodes.html#dgl.DGLGraph.batch_num_nodes), and for PyG, I think it's literally just `graph.batch`?

melo-gonzo: So would you prefer something like this?
`_, natoms = torch.unique(graph.batch, return_counts=True)`

laserkelvin: I think so, at least you can infer what's being done from the DGL equivalent as well :)

melo-gonzo: Ok, I updated it here [1c32408](https://github.com/IntelLabs/matsciml/pull/205/commits/1c32408aa7a6588f4615212152f43cdaf4d08830)

laserkelvin (approved):

PR #213, opened 2024-05-13
Prettifying PyTest output in Github actions

laserkelvin:
This PR adds `pytest-pretty` as a dependency to the actions workflows.

- Add `pytest-pretty` to `pyproject.toml` for `dev` specification
- Add `pytest-pretty` and a print formatting configuration to each PyTest workflow
- Configuring `pytest` to ignore deprecation warnings in `pyproject.toml`, which has been a huge (>300) source of text spam


melo-gonzo (approved):
🤩 this is great

PR #215, opened 2024-05-15
ASE `Calculator` interface

laserkelvin:
This PR adds a `Calculator` implementation that specializes to `matsciml`'s pipeline/abstractions.

Currently it is only implemented for energy and force prediction, but given that `ase` also supports other properties like stress, multipole moments, etc. there is some flexibility in using `ScalarRegressionTask` for things like that.

- Add `matsciml.interfaces.ase.MatSciMLCalculator` class that adapts `Atoms` representations to structures expected by `matsciml` models
- Add two minimum working examples for using the new interface: one loading from an inexistent pretrained checkpoint, and another showing training on LiPS and using `ase` after a quick training run on the devset
- Adds a Github action to run `pytest`s in the `interfaces` submodule
- Removes obsolete OCP relaxation and `ase` routines


matsciml/interfaces/ase/base.py
```diff
@@ -0,0 +1,201 @@
...
+        if len(self.results) == 0:
+            raise RuntimeError(
+                f"No expected properties were written. Output dict: {output}"
+            )
+
+    @classmethod
+    def from_pretrained_force_regression(
+        cls, ckpt_path: str | Path, *args, **kwargs
+    ) -> MatSciMLCalculator:
+        if isinstance(ckpt_path, str):
```
melo-gonzo, line 206: Can you consolidate this between the class methods?

matsciml/interfaces/ase/base.py
```diff
@@ -0,0 +1,201 @@
...
+            ckpt_path = Path(ckpt_path)
+        if not ckpt_path.exists():
+            raise FileNotFoundError(f"Checkpoint file not found; passed {ckpt_path}")
+        task = ForceRegressionTask.load_from_checkpoint(ckpt_path)
+        return cls(task, *args, **kwargs)
+
```
melo-gonzo, line 212: There should also be a `from_pretrained_multi_task` as it has a separate checkpoint loading style. We could either use `multitask_from_checkpoint` (in `matsciml/models/utils/io`), or write it up another way. Multitask does not work as it stands though, and even after loading through `multitask_from_checkpoint` the calculator pipeline will fail due to how the `MultiTaskLitModule` processes data.

On a related note, `MultiTaskLitModule` may have many output heads corresponding to an energy and force prediction, if more than one dataset was mapped to that task, so how should all of those outputs be selected for use in the calculator pipeline?

matsciml/interfaces/ase/base.py
```diff
@@ -0,0 +1,201 @@
...
+        return data_dict
+
+    def calculate(
+        self,
+        atoms=None,
+        properties: list[Literal["energy", "forces"]] = ["energy", "forces"],
+        system_changes=...,
+    ) -> None:
+        # retrieve atoms even if not passed
+        Calculator.calculate(self, atoms)
```
melo-gonzo, line 179: Is is possible to add logging to terminal here at any point? Not sure if ASE has a flag for that or not, but would be nice imo.

matsciml/models/base.py
```diff
@@ -1108,7 +1109,9 @@ def __init__(
     def _make_output_heads(self) -> nn.ModuleDict:
         modules = {}
         for key in self.task_keys:
-            modules[key] = OutputHead(1, **self.output_kwargs).to(self.device)
+            modules[key] = OutputHead(1, **self.output_kwargs).to(
+                self.device, dtype=self.dtype
```
melo-gonzo, line 1113: Mostly just curious what the need was for the `dtype` addition?

melo-gonzo (changes requested):
Looks good so far. Left a few comments. 

One more general comment is that it took me a while to figure out that using a checkpoint which was created from a task like this:
```python
model = EGNN(encoder_kwargs=...)
task = ForceRegressionTask(
    model,
    output_kwargs=...
)
```
will not load properly, however a checkpoint created from a task like this:
```python
task = ForceRegressionTask(
    encoder_class=EGNN,
    encoder_kwargs=...,
    output_kwargs=...
)
```
will load. This is due to how the 'hyper parameters' are stored in the model checkpoint.

It could be worth depreciating the first method of task creation in matsciml, or making note of this somewhere.

matsciml/interfaces/ase/base.py
laserkelvin: I was hoping to work on the multi task case in a separate PR, since it's a lot more complicated...

matsciml/interfaces/ase/base.py
laserkelvin: Not that I'm aware of - it only seems that you can log to file although you could technically jerry-rig it to write to a standard output stream (instead of a file one). If you really want this I would leave that to a separate PR

matsciml/interfaces/ase/base.py
melo-gonzo: This bit should be consolidated to one function:
```python
        if isinstance(ckpt_path, str):
            ckpt_path = Path(ckpt_path)
        if not ckpt_path.exists():
            raise FileNotFoundError(f"Checkpoint file not found; passed {ckpt_path}")
```

matsciml/interfaces/ase/base.py
melo-gonzo: That's fair, this one is pretty big as it stands.

matsciml/interfaces/ase/base.py
laserkelvin: Addressed in 439b7b4c

matsciml/interfaces/ase/base.py
laserkelvin: Raised #216 to track this.

melo-gonzo (approved):

PR #219, opened 2024-05-17
Checkpoint Path Check Bugfix

melo-gonzo:
The checkpoint check from #215 was not working as expected. Somthing about how `__` functions get tied up with the class in which they are associated with ([mangling](https://docs.python.org/3/tutorial/classes.html#private-variables)). This PR fixes this issue by making it a single underscore.

```python
  File "./matsciml/examples/interfaces/ase_from_pretrained.py", line 24, in <module>
    calc = MatSciMLCalculator.from_pretrained_force_regression(
  File "./matsciml/matsciml/interfaces/ase/base.py", line 264, in from_pretrained_force_regression
    ckpt_path = __checkpoint_conversion_exist_check(ckpt_path)
NameError: name '_MatSciMLCalculator__checkpoint_conversion_exist_check' is not defined
```


laserkelvin (approved):
Weird - I did the double underscore intentionally to make it private but didn't expect it to actually break things

PR #224, opened 2024-05-24
Multitask ASE interface

### LINKED ISSUE #217: [Feature request]: Reconciling multi task models with `ase` `Calculator` interface.
### Feature/behavior summary

Per #215 `MultiTaskLitModules` are a supported type of task to be passed into the calculator, but the `calculate` step doesn't treat it any differently even when it is actually required.

### Request attributes

- [ ] Would this be a refactor of existing code?

### Related issues

_No response_

### Solution description

We need a way to:

1. identify when inference results come from a `MultiTaskLitModule`
2. In the case where only one subtask is relevant, figure out how to extract those results and map them into the `Calculator`
3. In the case where _multiple_ subtasks are relevant, figure out how to aggregate the result and report a scalar as expected by `ase`

### Additional notes

_No response_

laserkelvin:
This PR closes #217 by providing an extended interface for mapping multitask models to `ase` workflows.

The main changes include:
- Introduces the `matsciml.interfaces.ase.multitask.AbstractStrategy` class, whose children allows us to implement ways to aggregate the outputs of multitasks.
- Implements the sole concrete class (for now), `AverageTasks`, which does a mean aggregation across the different dataset heads to compute energies/forces. Essentially, if you had `ForceRegressionTask` trained on multiple datasets, we outputs from all.
- Implements a custom `ase_calculate` method for `MultiTaskModule`, which effectively streamlines inference using multitask modules in scenarios like as an ASE calculator. It's named specifically for `ase`, but in the future we can use it as a dedicated inference method.

The `AbstractStrategy` should be flexible enough to do other potentially more intelligent aggregations, e.g. a weighted average, MoE approaches, etc.

### DIFF
--- a/examples/interfaces/ase_multitask_from_pretrained.py
+++ b/examples/interfaces/ase_multitask_from_pretrained.py
@@ -1,0 +1,42 @@
+from __future__ import annotations
+
+from ase import Atoms, units
+from ase.md.verlet import VelocityVerlet
+
+from matsciml.interfaces.ase import MatSciMLCalculator
+from matsciml.interfaces.ase.multitask import AverageTasks
+from matsciml.datasets.transforms import (
+    PeriodicPropertiesTransform,
+    PointCloudToGraphTransform,
+)
+
+"""
+Demonstrates setting up a calculator from a pretrained
+multitask/multidata module, using an averaging strategy to
+merge output heads.
+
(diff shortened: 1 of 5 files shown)


laserkelvin:
Ergh, issue with one of the tests...

PR #232, opened 2024-06-03
General predict method for tasks

laserkelvin:
This PR adds a base `predict` method to tasks, which is intended to be used for inference workflows. This method acts as a wrapper for the task `forward` call, but includes the additional step of applying the inverse normalization to the outputs, so that they are rescaled appropriately.

A summary of changes:

1. The `predict` definition in `BaseTaskModule`, which is inherited by all tasks except `MultiTaskModule`
2. Refactored the `MultiTaskModule.ase_calculate` to be `predict`, so that the interface is consistent. The workflow is more or less unchanged from before, but relies on `subtask.predict`.
3. For `ForceRegressionTask`, we override the `predict` logic slightly by rescaling both forces and node energies using the energy normalization factor _if_ dedicated normalizers are not provided by those values separately.
4. Updated the `ase` calculator code to rely on the `predict` interface, regardless of whether it is multitask or not.

### DIFF
--- a/matsciml/interfaces/ase/base.py
+++ b/matsciml/interfaces/ase/base.py
@@ -254,7 +254,8 @@ class MatSciMLCalculator(Calculator):
             results = self.multitask_strategy(output, self.task_module)
             self.results = results
         else:
-            output = self.task_module(data_dict)
+            # use the specialized predict method, which will rescale data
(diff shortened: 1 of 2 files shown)


melo-gonzo (approved):
Looks good!

PR #235, opened 2024-06-04
Optional Task Loss Scaling

melo-gonzo:
This PR adds the ability to optionally add a multiplicative scale factor to the loss value for each task head (`scaled_loss = loss * scale_value`). This parameter is an optional input to the specific task being used which maps specified loss scaling keys to existing task keys. Any task keys that do not have a task_loss_scaling key will default to 1. 

Example:
```python
task = ScalarRegressionTask(
  **egnn_config,
  task_keys=["efermi", "energy_total", "relative_energy"],
  task_loss_scaling={"efermi": 1e-5, "energy_total": 1e4},
)
```
In the above example, the two keys `efermi` and `energy_total` will have a specific scale applied, whicle `relative_energy` will default to a scale factor of 1.

### DIFF
--- a/matsciml/models/base.py
+++ b/matsciml/models/base.py
@@ -698,6 +698,8 @@ class BaseTaskModule(pl.LightningModule):
         self.output_kwargs = default_heads
         self.normalize_kwargs = normalize_kwargs
         self.task_keys = task_keys
+        breakpoint()
+        self.task_loss_scaling = None
         self.embedding_reduction_type = embedding_reduction_type
         self.save_hyperparameters(ignore=["encoder", "loss_func"])
 
@@ -741,6 +743,31 @@ class BaseTaskModule(pl.LightningModule):
         # some tasks like ForceRegressionTask doesn't actually use an output
         # head for the forces
         return True
+
+    @property
+    def task_loss_scaling(self) -> list[str]:
+        return self._task_loss_scaling
+
+    @task_loss_scaling.setter
+    def task_loss_scaling(self, loss_scaling_dict: dict[str, float] | None) -> None:
+        """
+        Adds loss scaling per task. Ensures task loss scaling key is present in the 
+        task loss keys.
+
+        Parameters
+        ----------
+        loss_scaling_dict : dict[str, float]
+            Dictionary used to map a task key to a scaling value.
+        """
+
+        # Check if task loss scaling key is present in task keys.
+
+        if loss_scaling_dict is None:
+            loss_scaling_dict = {}
+        if not isinstance(loss_scaling_dict, dict):
+            raise TypeError(f"task_loss_scaling {loss_scaling_dict} must be a dictionary, got type {type(loss_scaling_dict)}.")
+        self._task_loss_scaling = loss_scaling_dict
+        self.hparams["task_loss_scaling"] = self._task_loss_scaling
 
     @abstractmethod
     def _make_output_heads(self) -> nn.ModuleDict: ...
--- a/matsciml/models/base.py
+++ b/matsciml/models/base.py
@@ -746,7 +746,7 @@ class BaseTaskModule(pl.LightningModule):
         return True
 
     @property
-    def task_loss_scaling(self) -> list[str]:
+    def task_loss_scaling(self) -> dict[str, float]:
         return self._task_loss_scaling
 
     @task_loss_scaling.setter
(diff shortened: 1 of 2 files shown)


matsciml/models/base.py
```diff
@@ -698,6 +698,9 @@ def __init__(
         self.output_kwargs = default_heads
         self.normalize_kwargs = normalize_kwargs
         self.task_keys = task_keys
+        self._task_loss_scaling = kwargs.get("task_loss_scaling", {})
+        if self.task_keys != []:
```
laserkelvin, line 702: So basically we only set the scaling if task keys are passed specifically? I might prefer `len(self.task_keys) > 0`

matsciml/models/base.py
```diff
@@ -698,6 +698,9 @@ def __init__(
         self.output_kwargs = default_heads
         self.normalize_kwargs = normalize_kwargs
         self.task_keys = task_keys
+        self._task_loss_scaling = kwargs.get("task_loss_scaling", {})
```
laserkelvin, line 701: The default value is a `dict`, but your property implementation hints at a list of strings. I think the type hint is wrong?

matsciml/models/base.py
```diff
@@ -742,6 +745,42 @@ def has_initialized(self) -> bool:
         # head for the forces
         return True
 
+    @property
+    def task_loss_scaling(self) -> list[str]:
```
laserkelvin, line 749: Aforementioned type hint correction

matsciml/models/base.py
```diff
@@ -742,6 +745,42 @@ def has_initialized(self) -> bool:
...
+        """
+        if loss_scaling_dict is None:
+            loss_scaling_dict = {1: key for key in self._task_keys}
+
+        if not isinstance(loss_scaling_dict, dict):
+            raise TypeError(
+                f"task_loss_scaling {loss_scaling_dict} must be a dictionary, got type {type(loss_scaling_dict)}."
+            )
+
+        for scaling_key in loss_scaling_dict.keys():
```
laserkelvin, line 771: This is more personal style, but not sure if it's worth merging the two key sets together and just working on a single loop to make it look cleaner.

e.g.

```python
keys = set(loss_scaling_dict.keys()).intersection(set(self._task_keys))
not_in_scaling = []
not_in_tasks = []
final_scaling_dict = {}
for key in keys:
    if key not in loss_scaling_dict:
        not_in_scaling.append(key)
    if key not in self._task_keys:
        not_in_tasks.append(keys)
    final_scaling_dict[key] = loss_scaling_dict.get(key, 1.0)

if len(not_in_scaling) != 0:
    warn(f"Unspecified tasks will have a scaling of 1: {not_in_scaling")
if len(not_in_tasks) != 0:
    warn(f"Task scaling keys passed that are not actually tasks: {not_in_tasks}
```

IMO it's easier to read, and also if you have a bunch of keys that don't belong or not found you won't get spammed by a bunch of warning messages.

laserkelvin (changes requested):

matsciml/models/base.py
melo-gonzo: This check is used if no task keys are sent in, but rather they get created in the `on_train_batch_start` call. I agree that `len(self.task_keys) > 0` looks cleaner. Updated in [2428b2b](https://github.com/IntelLabs/matsciml/pull/235/commits/2428b2bf8692069201e2b3d0e7efa15a9b5bd52d)

matsciml/models/base.py
melo-gonzo: Updated in [61e3597](https://github.com/IntelLabs/matsciml/pull/235/commits/61e3597bfd9498d01447a6e3d06196767200a174)

matsciml/models/base.py
melo-gonzo: Good eye - updated in [61e3597](https://github.com/IntelLabs/matsciml/pull/235/commits/61e3597bfd9498d01447a6e3d06196767200a174)

matsciml/models/base.py
melo-gonzo: I've got to disagree with the readability, but I agree with the spamming of warnings. I've updated to your suggestion, with a minor change, replacing intersection with union. Updated here [37509bd](https://github.com/IntelLabs/matsciml/pull/235/commits/37509bde88ca8d4f73e2036b4c150aa9dc5087b7)

laserkelvin: If you disagreed on readability you didn't have to make the change 😆

laserkelvin (approved):
LGTM

PR #237, opened 2024-06-07
Quality of life and helper callback functions

laserkelvin:
This PR introduces and adds a bunch of changes pertaining to informing the user of things happening under the hood, particularly during training.

One of the big philosophical changes is also focusing more on enabling logging to be done with `TensorBoardLogger` and `WandbLogger`s by writing functions more tailored to them, rather than before where loggers were treated in the abstract entirely.

## Summary

- Changed the use of coordinates in periodic boundary utilities to use cartesian coordinates, not fractional coordinates. Also included a warning message that looks at the coordinates as part of diagnostics.
- For model training, task modules now include `log_embeddings` and `log_embeddings_every_n_steps` arguments that are saved to `hparams`, which as the pair suggests, allow you to regularly log embedding vectors for analysis. This will let you ensure oversmoothing doesn't occur, where all of the embedding features become identical.
- Introduced a `TrainingHelperCallback`, which is intended to help diagnose some common issues with training, such as unused parameters, missing gradients, tiny gradients, etc. Complimentary to the change above, there is an option to inject a forward hook to any encoder (assuming it produces an `Embeddings` structure), and uses it to calculate the variance in embeddings.
- Introduced a `ModelAutocorrelation` callback, which will perform an autocorrelation analysis on model parameters and gradients over the course of training. Basically this gives you some insight into how the training dynamics appear, i.e. too much correlation = probably not good.

My intention for the `TrainingHelperCallback` is to be like a guide for best practices: we can refine this as we go and discover new things, and hopefully will be useful for everyone including new users.

### DIFF
--- a/examples/callbacks/autocorrelation.py
+++ b/examples/callbacks/autocorrelation.py
@@ -1,0 +1,65 @@
+from __future__ import annotations
+
+import pytorch_lightning as pl
+from pytorch_lightning.loggers import TensorBoardLogger, WandbLogger
+
+from matsciml.datasets.transforms import DistancesTransform, PointCloudToGraphTransform
+from matsciml.lightning.data_utils import MatSciMLDataModule
+from matsciml.lightning.callbacks import ModelAutocorrelation
+from matsciml.models import SchNet
+from matsciml.models.base import ScalarRegressionTask
+
+"""
+This script demonstrates the use of the `ModelAutocorrelation` callback.
+
+The main utility of this callback is to monitor the degree of correlation
+in model parameters and optionally gradients over a time span. The idea
+is that for optimization trajectories, steps are ideally as de-correlated
+as possible (at least within reason), and indeed is actually a major
+assumption of Adam-like optimizers.
+
+There is no hard coded heuristic for identifying "too much correlation"
+yet, however this callback can help do the data collection for you to
+develop a sense for yourself. One method for trying this out is to
+set varying learning rates, and seeing how the autocorrelation spectra
+are different.
+"""
+
+# construct a scalar regression task with SchNet encoder
+task = ScalarRegressionTask(
+    encoder_class=SchNet,
+    # kwargs to be passed into the creation of SchNet model
+    encoder_kwargs={
+        "encoder_only": True,
+        "hidden_feats": [128, 128, 128],
+        "atom_embedding_dim": 128,
+    },
+    # which keys to use as targets
+    task_keys=["energy_relaxed"],
+    log_embeddings=False,
+)
+# Use IS2RE devset to test workflow
+# SchNet uses RBFs, and expects edge features corresponding to atom-atom distances
+dm = MatSciMLDataModule.from_devset(
+    "IS2REDataset",
+    dset_kwargs={
+        "transforms": [
+            PointCloudToGraphTransform(
+                "dgl",
+                cutoff_dist=20.0,
+                node_keys=["pos", "atomic_numbers"],
+            ),
+            DistancesTransform(),
+        ],
+    },
+)
+
+# tensorboard logging if working purely locally, otherwise wandb
+logger = WandbLogger(
+    name="helper-callback", offline=False, project="matsciml", log_model="all"
+)
+logger = TensorBoardLogger("./")
+
+# run a quick training loop
+trainer = pl.Trainer(max_epochs=30, logger=logger, callbacks=[ModelAutocorrelation()])
+trainer.fit(task, datamodule=dm)
--- a/matsciml/lightning/callbacks.py
+++ b/matsciml/lightning/callbacks.py
@@ -884,7 +884,7 @@ def embedding_magnitude_hook(
                 logger.warning(
                     f"Median system/graph embedding value is greater than 10 ({sys_z_med})"
                 )
-            if sys_z_var <= 1e-2:
+            if sys_z_var <= 1e-5:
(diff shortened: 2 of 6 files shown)


laserkelvin:
I think I have a lead on what the issue is: because of how `SAM` works, and because of the modifications to "stashing" embeddings in the batch structure, we now end up with two disjoint computational graphs that causes `backward` to break.

This needs a bit of thought to fix...

laserkelvin:
Confirming this by changing out the `BaseTaskModule.forward`:

```python
        if "embeddings" in batch:
            embeddings = batch.get("embeddings")
        else:
            embeddings = self.encoder(batch)
            batch["embeddings"] = embeddings
        outputs = self.process_embedding(embeddings)
        return outputs
```

Removing the branch, and just running the encoder + processing embeddings works (i.e. don't try and grab cached embeddings).

Ideally there would be a way to check if embeddings originated from the same computational graph, but that take a lot more surgery than this PR warrants. I'll think of an alternative to this.

matsciml/lightning/callbacks.py
```diff
@@ -845,3 +850,563 @@ def _second_step(
...
+        for parameter in pl_module.encoder.parameters():
+            encoder_norm_vals.append(parameter.detach().norm().cpu().item())
+        for head in pl_module.output_heads.values():
+            for parameter in head.parameters():
+                output_norm_vals.append(parameter.detach().norm().cpu().item())
+        encoder_norm_vals = np.array(encoder_norm_vals)
+        output_norm_vals = np.array(output_norm_vals)
+        encoder_median = np.median(encoder_norm_vals)
+        output_median = np.median(output_norm_vals)
+        if encoder_median < (2.0 * output_median):
```
melo-gonzo, line 1053: Another param that should probably be set in the module initialization.

matsciml/lightning/callbacks.py
```diff
@@ -845,3 +850,563 @@ def _second_step(
...
+            python_logger.warning(
+                "Median encoder weights are significantly smaller than output heads:"
+                " encoder median norm: {encoder_median:.3e},"
+                " output head: {output_median:.3e}"
+            )
+        # optionally record to service as well
+        if log_history and pl_module.logger is not None:
+            log_service = pl_module.logger.experiment
```
melo-gonzo, line 1061: Probably not a huge deal, but when multiple loggers are used this will default to the first logger entered, for example loggers=[csv_logger, wandb_logger], `pl_module.logger` would be csv_logger in this case. Again, probably unlikely to encounter this in the wild - just making note. (I use csv logger + wandb 🙂). [loggers code](https://github.com/Lightning-AI/pytorch-lightning/blob/06ea3a05716a6d1f4a96cfb25021accdd18d8146/src/lightning/pytorch/trainer/trainer.py#L1575)

melo-gonzo (commented):
Some great features here, thanks for doing all of this! I threw in a few comments, I know you're still working on things.

matsciml/lightning/callbacks.py
```diff
@@ -845,3 +850,563 @@ def _second_step(
...
+    logger = getLogger("matsciml.helper")
+    logger.setLevel("INFO")
+    if isinstance(output, Embeddings):
+        # check the magnitude of both node and system level embeddings
+        if output.system_embedding is not None:
+            sys_z = output.system_embedding.detach().cpu()
+            # calculate representative statistics
+            sys_z_med = sys_z.median().item()
+            sys_z_var = sys_z.var().item()
+            if sys_z_med > 10.0:
```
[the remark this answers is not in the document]
laserkelvin: The number is based my own heuristics

Because this is meant to be a function registered as a forward hook, I can't really play with the arguments too easily. In principle we could make the function dynamically generated, but I don't have the bandwidth right now to refactor it...

matsciml/lightning/callbacks.py
```diff
@@ -845,3 +850,563 @@ def _second_step(
...
+                logger.warning(
+                    f"Median system/graph embedding value is greater than 10 ({sys_z_med})"
+                )
+            if sys_z_var <= 1e-5:
```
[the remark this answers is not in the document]
laserkelvin: See my comment above, same applies here

examples/callbacks/autocorrelation.py
```diff
@@ -0,0 +1,65 @@
...
+                cutoff_dist=20.0,
+                node_keys=["pos", "atomic_numbers"],
+            ),
+            DistancesTransform(),
+        ],
+    },
+)
+
+# tensorboard logging if working purely locally, otherwise wandb
+logger = WandbLogger(
```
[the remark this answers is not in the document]
laserkelvin: I guess strictly speaking no, since these are optional functions. If a user wants to use these functions, they can install one or both (I don't want to pin them right now)

matsciml/lightning/callbacks.py
laserkelvin: This one again is like another hardcoded heuristic, similar reason but different since this is a `staticmethod` we don't have access to the callback's attributes unless we refactor heavily

matsciml/lightning/callbacks.py
laserkelvin: Thanks for pointing that out! Addressed in aeab6cdf

melo-gonzo (approved):
This will bring some great utilities and helpful debugging tools! Looks good to merge.

PR #240, opened 2024-06-10
Manual step for learning rate schedulers in `ForceRegressionTask`

laserkelvin:
This PR fixes the unintended behavior of not stepping learning rate schedulers during `ForceRegressionTask` training. Because we use manual optimization, this kind of flew under the radar and I caught it while monitoring the learning rate.

The way this is implemented assumes the learning rate is stepped once per epoch: in some cases a per-batch cadence might be desirable, but can cross that bridge when we get there.

### DIFF
--- a/matsciml/models/base.py
+++ b/matsciml/models/base.py
@@ -1944,6 +1944,21 @@ class ForceRegressionTask(BaseTaskModule):
             )
             batch_size = None
         self.log_dict(metrics, on_step=True, prog_bar=True, batch_size=batch_size)
+        # step learning rate schedulers at the end of epochs
+        if self.trainer.is_last_batch:
+            schedulers = self.lr_schedulers()
+            if isinstance(schedulers, list):
+                for s in schedulers:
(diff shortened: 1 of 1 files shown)


matsciml/models/base.py
```diff
@@ -1944,6 +1944,21 @@ def training_step(
             )
             batch_size = None
         self.log_dict(metrics, on_step=True, prog_bar=True, batch_size=batch_size)
+        # step learning rate schedulers at the end of epochs
+        if self.trainer.is_last_batch:
+            schedulers = self.lr_schedulers()
+            if isinstance(schedulers, list):
```
melo-gonzo, line 1950: You could just handle this as a list every time and not repeat the code block
```python
if not isinstance(schedulers, list):
    schedulers = [schedulers]
```

laserkelvin: Addressed in 203a3d9

melo-gonzo (approved):

PR #228, opened 2024-05-28
Rename src/dst Nodes For Point Cloud Featurization

melo-gonzo:
Would close the issue raised in #132 and #227. Renames the `src_nodes` and `dst_nodes` that are used for point cloud featurization to `pc_src_nodes` and `pc_dst_nodes`.

PR #246, opened 2024-06-27
Experiments Running and Management

melo-gonzo:
This PR adds functionality to run and manage experiments in matsciml. Currently, the options for experimentation are to create a script for each experiment run and manually update parameters, or to use the pytorch lightning cli + yaml files. The first approach is time-consuming and clumsy, and the second approach does not handle multi task and multi data experiments in matsciml. This PR aims to bridge the gap between functional experimental pipelines without the headache of managing a ton of scripts. It takes inspiration from the lightning cli by managing the different aspects of an experiment with yaml files `{trainer, model, dataset, experiment}.yaml` and allows for cli updates by specifying a chain of parameters to traverse and update. Experimentation effectively comes down to managing these yaml files and greatly reduces the barrier to entry towards running complex experiments with multi task and multi data origins. See the README for more details.


melo-gonzo:
Note that I updated `matsciml/models/dgl/gaanet/tests/test_gala.py` and `matsciml/models/base.py` to match changes from #228.

experiments/datasets/data_module_config.py
laserkelvin, line 1: Do you want to move this things into the main datasets module?

You could repackage these functions as a `from_config` class method, then you don't need to include `__init__.py` in this folder. It _could_ mess up `pip install`s

experiments/README.md
```diff
@@ -0,0 +1,127 @@
+# Experiments in Open MatSciML Toolkit
+
+Experimental workflows may be time consuming, repetitive, and complex to set up. Additionally, pytorch-lightning based cli utilities may not be able to handle specific use cases such as multi-data or multi-task training in matsciml. The experiments module of MatSciML is meant to loosely mirror the functionality of the pytorch lightning cli while allowing more flexibility in setting up complex experiments. Yaml files define the module parameters, and specific arguments may be change via the command line if desired. A single command is used to launch training runs which take out the complexity of writing up new script for each experiment type.
+
+## Experiment Config
+The starting point of defining an experiment is the experiment config. This is a yaml file that lays out what model, dataset(s), and task(s) will be used during training. An example config for single task training yaml (`single_tasl.yaml`) look like this:
```
laserkelvin, line 6: Typo on `single_tasl.yaml`

experiments/datasets/materials_project.yaml
```diff
@@ -0,0 +1,20 @@
+dataset: MaterialsProjectDataset
+debug:
+  batch_size: 4
+  num_workers: 0
+experiment:
+  test_split: ./matsciml/datasets/materiasl_project/devset
```
laserkelvin, line 6: Is this intentional? I don't think it would point to anything

experiments/task_config/task_config.py
```diff
@@ -0,0 +1,61 @@
...
+    MultiTaskLitModule,
+    ScalarRegressionTask,
+)
+
+from experiments.datasets import available_data
+from experiments.models import available_models
+from experiments.utils.utils import instantiate_arg_dict, update_arg_dict
+
+
+task_map = {
```
laserkelvin, line 23: Would you like to have it so that each task can have a private variable `__alias__` that we could map to?

e.g.:

```python
class ForceRegressionTask:
    __alias__ = "fr"

    def __init__(self, ...):
```

We could then generate the mapping in `models.base`:

```python
# in models/base.py

alias_map = {}
for key in __all__:
    # not sure if this works, but something to this effect
    alias_map[key] = getattr(__file__, key)
```

Alternatively, we could actually use the registry?

```python
from matsciml.common.registry import registry

# once again in models/base.py
for key in __all__:
    registry.register_model(getattr(__file__, key), key)

# and then use registry in `setup_task`
```

laserkelvin (changes requested):
I've put in some minor comments, but the two high level things that I'll bring up for discussion are:

1. Maybe we can _finally_ use the registry, so instead of maintaining individual maps in task, dataset, and models, just rely on `matsciml.common.registry` to do the mapping?
2. I'm not sure if I missed it, but it might be good to refactor your object creation functions (i.e. mapping `args` into something you pull out of a mapping) into `matsciml.common.registry`. The "improvement" so to speak would be to also include `arg`/`kwarg` validation, similar to [this](https://github.com/IntelLabs/matsciml/blob/430712f57af820fc918538324cb8c3f2b1947795/matsciml/models/common.py#L210) and I think also in `MACEWrapper`. I think you want to raise an exception if someone passes something that isn't recognized, and this would be a good place to put it (even as a `staticmethod` of `Registry`).

experiments/README.md
melo-gonzo: fixed [81b8e06](https://github.com/IntelLabs/matsciml/pull/246/commits/81b8e060ffc66a2ecba6ded9a7c518f7018b1921)

experiments/datasets/materials_project.yaml
melo-gonzo: fixed [2bad280](https://github.com/IntelLabs/matsciml/pull/246/commits/2bad280616fb9f72aa3045662dc502e37367bfa1)

experiments/task_config/task_config.py
melo-gonzo: I went ahead and purged the 'task_map', and opting to just specify the task class name and using the registry to get out the actual class object. In [this](https://github.com/IntelLabs/matsciml/pull/246/commits/f4364032b4ab8b69b82d40e3ea7ff70ed56ec106) commit

melo-gonzo:
> bb140fb

I added the class argument verification check in [this](https://github.com/IntelLabs/matsciml/pull/246/commits/bb140fbc92f06a2526da62fdf8ba270ad0ce7e8f) commit. I'm not sure it makes sense to refactor the object creation function `instantiate_arg_dict` and the argument checker `verify_class_args` into the `Registry` class, because these functions also are used to spin up lightning classes such as callbacks and loggers, and other random functions that may not be a part of the registered objects, such as matsciml.datasets.utils.element_types`, and lots of the MACE moules.

experiments/datasets/data_module_config.py
melo-gonzo: I think I'd prefer to keep this separate from the datasets module since it is also relying on other experiment related utils such as the command line parsing. If there ends up being other use cases where this functionality would be helpful from the dataset module we can do a refactor then.

experiments/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+In general, and experiment may the be launched by running:
+`python experiments/training_script.py --experiment_config ./experiments/single_task.yaml`
+
+
+* The `model` field points to a specify `model.yaml` file in `./experiments/models`.
+* The `dataset` field is a dictionary specifying which datasets to use, as well as which tasks are associated with the parent dataset.
+    * Tasks are referred to by a shortened version of their name:
+    ```python
+    "sr","scalar_regression" -> ScalarRegressionTask
```
laserkelvin, line 23: So this shortened version won't work anymore, right?

experiments/datasets/__init__.py
```diff
@@ -0,0 +1,20 @@
+import os
+import yaml
+
+
+yaml_dir = yaml_dir = os.path.dirname(os.path.abspath(__file__))
```
laserkelvin, line 5: I'll make this optional for you, but just looks cleaner if you use `pathlib` instead:

```python
from pathlib import Path
yaml_dir = Path(__file__)

for filename in yaml_dir.rglob("*.yaml"):
    ...
```

experiments/datasets/tests/test_data_module_creation.py
```diff
@@ -0,0 +1,24 @@
...
+
+import matsciml
+import matsciml.datasets.transforms  # noqa: F401
+from experiments.datasets.data_module_config import setup_datamodule
+
+single_task = {"dataset": {"oqmd": {"sr": ["energy"]}}}
+multi_task = {"dataset": {"s2ef": {"sr": "energy", "fr": ["force"]}}}
+multi_data = {
+    "dataset": {
+        "s2ef": {"sr": "energy", "fr": ["force"]},
```
laserkelvin, line 14: Do the abbreviations still work?

experiments/models/__init__.py
```diff
@@ -0,0 +1,28 @@
+import os
+import yaml
+
+from torch.nn import LayerNorm
+
+
+yaml_dir = yaml_dir = os.path.dirname(os.path.abspath(__file__))
```
laserkelvin, line 7: Same comment as for `datasets`, use `pathlib` instead of `os.path`

experiments/trainer_config/__init__.py
```diff
@@ -0,0 +1,18 @@
+from experiments.trainer_config.trainer_config import setup_trainer  # noqa: F401
+
+import os
+import yaml
+
+
+yaml_dir = yaml_dir = os.path.dirname(os.path.abspath(__file__))
```
laserkelvin, line 7: Same comment for `pathlib`

experiments/training_script.py
```diff
@@ -0,0 +1,41 @@
...
+
+    dm = setup_datamodule(config)
+    task = setup_task(config)
+    trainer = setup_trainer(config, trainer_args=trainer_args)
+    trainer.fit(task, datamodule=dm)
+
+
+if __name__ == "__main__":
+    parser = ArgumentParser()
+    parser.add_argument("-o", "--options", action="store_true")
```
laserkelvin, line 26: Could you add `help=...` for each of these arguments? I imagine running `python training_script.py --help` would be a common theme

matsciml/models/base.py
```diff
@@ -337,12 +337,18 @@ def read_batch(self, batch: BatchDict) -> DataDict:
...
+                batch["pc_src_nodes"][index],
+                batch["pc_dst_nodes"][index],
+            )
             # use dst_nodes to gauge size because you will always have more
             # dst nodes than src nodes right now
             sizes.append(len(dst_nodes))
             # carve out neighborhoods as dictated by the dataset/transform definition
-            sample_pc_pos = sample[src_nodes][None, :] - sample[dst_nodes][:, None]
+            sample_pc_pos = (
+                sample[src_nodes.to(int)][None, :]
```
laserkelvin, line 349: I think you want to do `src_nodes.long()` and `dst_nodes.long()` instead, since it's less ambiguous.

laserkelvin (changes requested):

experiments/README.md
melo-gonzo: fixed [759a4fc](https://github.com/IntelLabs/matsciml/pull/246/commits/759a4fc94a377e49141fbab531b3fb393d7af02f)

experiments/datasets/__init__.py
melo-gonzo: fixed [759a4fc](https://github.com/IntelLabs/matsciml/pull/246/commits/759a4fc94a377e49141fbab531b3fb393d7af02f)

experiments/datasets/tests/test_data_module_creation.py
melo-gonzo: fixed [759a4fc](https://github.com/IntelLabs/matsciml/pull/246/commits/759a4fc94a377e49141fbab531b3fb393d7af02f)

experiments/models/__init__.py
melo-gonzo: fixed [759a4fc](https://github.com/IntelLabs/matsciml/pull/246/commits/759a4fc94a377e49141fbab531b3fb393d7af02f)

experiments/trainer_config/__init__.py
melo-gonzo: fixed [759a4fc](https://github.com/IntelLabs/matsciml/pull/246/commits/759a4fc94a377e49141fbab531b3fb393d7af02f)

experiments/training_script.py
melo-gonzo: fixed [35e44e9](https://github.com/IntelLabs/matsciml/pull/246/commits/35e44e9d4cd3b530952009a96c4a25dfba27c97d)

matsciml/models/base.py
melo-gonzo: fixed [8e11437](https://github.com/IntelLabs/matsciml/pull/246/commits/8e1143709dd38cf49b4a275c8685136fe25931eb)

laserkelvin (approved):
Thanks for making all the changes, LGTM!

PR #248, opened 2024-07-02
Default MACE atomic energies

laserkelvin:
This PR replaces the default MACE atomic energies dictionary with the total free ion energy, instead of just a tensor of ones. Arguably this is more principled, but makes the default configuration a little more opinionated.


melo-gonzo (approved):
Good to merge, pending one small update in the logger.

PR #257, opened 2024-07-19
Implementing exponential moving average Lightning callback

laserkelvin:
This PR adds an exponential moving average callback for improved generalization in training dynamics, in a modular fashion.

The callback works entirely independently of a given task: the EMA model updates are performed by this callback's training hooks, rather than by the task itself. That said, base methods in tasks were updated such that inference is done with the EMA model weights if they are available.

An example script is also provided showing how to use both SWA and EMA callbacks in a training run.


laserkelvin:
The SAM callback tests are failing - don't think it's due to things I've changed...

melo-gonzo (approved):
Looks good, this will be nice to try on our supported architectures!

PR #262, opened 2024-07-25
Refactor Experiment Configs

melo-gonzo:
This PR refactors the experiment configs. Previously, the config files needed to be present in their respective directories, making it hard to separate out config files for different groups of experiments. This PR moves all the default configs into a folder called `configs` with subfolders for each config type. Now, the user may create folders for their own specific experiments, use a combination of default config files and custom configs by specifying a path to each respective config in the cli arguments.


laserkelvin (approved):
Thanks for the heavy refactor! Our quality of life will shoot through the roof now 🚀

PR #266, opened 2024-08-05
Implementing `MaterialsTrajectoryDataset`

laserkelvin:
This PR adds a subclass of Materials Project as the `MaterialsTrajectoryDataset`, which focuses specifically on energy/force/stress regression tasks.

For the energy, we use the corrected total energy key, allowing the dataset to be used with `ForceRegressionTask` without further modification.


melo-gonzo (approved):
Seems like some of the pytests are unhappy, otherwise LGTM!

laserkelvin:
Commit 99a90d9 adds a more informative exception handling. The pytests fail because we don't have a devset for MPTraj yet

PR #270, opened 2024-08-08
Implementing loss scaling scheduler callback and schedulers

laserkelvin:
This PR adds a new callback, called `LossScalingScheduler`, and scheduler classes that will modify the relative loss weights over the course of training.

- Two types of schedulers are implemented: `LinearScalingSchedule` and `SigmoidScalingSchedule`. The former will generate a linear ramp from start to end over steps or epochs, and the latter gives a gradual ramp up in the form of a sigmoid curve.
- The `LossScalingScheduler` is configured by passing schedules that are mapped to task keys, and for every training step or epoch (set by the schedules), applies a new value of the weighting to the appropriate task.
- The example script `examples/callbacks/loss_scheduling.py` is provided to show how it is configured.

This is useful for implementing curricula, where we prioritize learning of different properties over time.

### DIFF
--- a/matsciml/lightning/callbacks.py
+++ b/matsciml/lightning/callbacks.py
@@ -1496,7 +1496,7 @@ class ExponentialMovingAverageCallback(Callback):
 
 
 class LossScalingScheduler(Callback):
-    def __init__(self, *schedules: list[BaseScalingSchedule]) -> None:
+    def __init__(self, *schedules: BaseScalingSchedule) -> None:
         """
         Callback for dynamically adjusting loss scaling values over
         the course of training, a la curriculum learning.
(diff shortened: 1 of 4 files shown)


laserkelvin:
I think the same tests fail as in #266

matsciml/lightning/callbacks.py
```diff
@@ -1492,3 +1511,95 @@ def on_fit_end(
...
+                try:
+                    new_scaling_value = schedule.step()
+                    pl_module.task_loss_scaling[target_key] = new_scaling_value
+                    self._logger.debug(
+                        f"Advanced {target_key} to new value: {new_scaling_value}"
+                    )
+                except StopIteration:
+                    self._logger.warning(
+                        f"{target_key} has run out of scheduled values; this may be unintentional."
+                    )
```
melo-gonzo, line 1585: Looks like you could combine this common bit between the on_x_end functions, and pass in a 'step' or 'epoch' string variable to use in the if statement and log message.

melo-gonzo (approved):
Just one general comment, otherwise looks good!

matsciml/lightning/callbacks.py
laserkelvin: Done in d3b97da

PR #271, opened 2024-08-12
Noisy node modification

laserkelvin:
This PR adds a flag to rescale the noise targets for noisy nodes prediction, as per Liao 2024. We also modify the default scale value to be much larger.

### DIFF
--- a/matsciml/datasets/transforms/pretraining/noisy_positions.py
+++ b/matsciml/datasets/transforms/pretraining/noisy_positions.py
@@ -11,7 +11,7 @@ __all__ = ["NoisyPositions"]
 
 
 class NoisyPositions(AbstractDataTransform):
-    def __init__(self, scale: float = 1e-3) -> None:
+    def __init__(self, scale: float = 1e-3, normalize: bool = False) -> None:
         """
         Initializes a NoisyPositions transform.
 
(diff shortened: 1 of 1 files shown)


matsciml/datasets/transforms/pretraining/noisy_positions.py
```diff
@@ -58,6 +62,9 @@ def __call__(self, data: DataDict) -> DataDict:
         else:
             data["noisy_pos"] = noisy_pos
         # set targets so that tasks know what to do
+        # noise targets are normalized by the scale
+        if self.normalize:
+            noise /= self.scale
```
melo-gonzo, line 67: Is this not the same as just using `scale=1` since `noise = random * scale`? I may be missing the point

laserkelvin: Yeah - it's basically making the model predict noise that doesn't depend on the scale. Since we're not actually using the predicted noise in any modeling capacity (e.g. in a diffusion model) this is just to make the number behave predictably.

To be clear the node positions are shifted by noise with `scale != 1`, but the _loss target_ is normalized.

melo-gonzo: I see, thanks for the clarification.

melo-gonzo (approved):

Issue #263, opened 2024-07-29
[Feature request]: Allow weights transfer and/or restarting from earlier checkpoint with experiment CLI


laserkelvin:
### Feature/behavior summary

To my knowledge, the experiment CLI doesn't support `load_from_checkpoint` or `from_pretrained_encoder` type of restarts, which would be very useful for either restarting training (with more epochs, say), or to transfer an encoder to a new task.

### Request attributes

- [X] Would this be a refactor of existing code?
- [ ] Does this proposal require new package dependencies?
- [ ] Would this change break backwards compatibility?
- [ ] Does this proposal include a new model?
- [ ] Does this proposal include a new dataset?
- [ ] Does this proposal include a new task/workflow?

### Solution description

An example solution would be including key/value pairs in a model/task YAML configuration that will let you set the type of reload a user would like to, and where the weights are located.

For example for continuing training:

```yaml
load_weights:
   method: checkpoint
   type: local
   path: ./path/to/lightning_checkpoint.ckpt
```

To transfer the weights from a pretrained encoder, _and_ load weights from a `wandb` artifact:

```yaml
load_weights:
   method: pretrained
   type: wandb
   path: "<wandb_user>/<project>/<model_name>:v<version>"
```

The `method` key would determine whether you would use `Task.load_from_checkpoint` or `Task.from_pretrained_encoder` methods. The `type` can be used to indicate the kind of checkpoint, i.e. a local file or from `wandb`. In the latter case, it would be a little bit more complicated as you would have would have use the `wandb` API to initialize a run that doesn't synchronize, download the weight artifact, _then_ map to a load method.

An alternative interface could be to just pass those method names directly:

```yaml
load_from_checkpoint: "local:./path/to/lightning_checkpoint.ckpt"
```

and

```yaml
from_pretrained_encoder: "wandb:<wandb_user>/<project>/<model_name>:v<version>"
```

PR #272, opened 2024-08-14
Allow Loading of Checkpoints in Experiments Pipeline

melo-gonzo:
This PR addressed #263 for loading checkpoints using the experiments cli. Local or WandB stored checkpoints may be loaded and used for continual training, or encoder weight transfer to a new task. Checkpoints associated with training from the `MultiTaskLitModule` are processed using `matsciml.models.multitask_from_checkpoint` which has limited functionality at the moment, and is only suited for continuing training.

### DIFF
--- a/experiments/README.md
+++ b/experiments/README.md
@@ -21,6 +21,27 @@ dataset:
       - energy
 ```
 
+### Checkpoint Loading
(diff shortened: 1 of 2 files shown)


laserkelvin (approved):
Looks fantastic!

Issue #275, opened 2024-08-19
[Bug]: ase calculator breaks when using MACE checkpoint


lory-w:
### Expected behavior

Load MACE checkpoint as an ase calculator and calculate atoms potential energy 

### Actual behavior

```
  File "/store/code/matsciml/matsciml/models/base.py", line 292, in forward
    outputs = self._forward(**input_data)
  File "/store/code/matsciml/matsciml/models/pyg/mace/wrapper/model.py", line 206, in _forward
    outputs = self.encoder(
  File "/store/.local/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1553, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/store/.local/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1562, in _call_impl
    return forward_call(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/mace/modules/models.py", line 339, in forward
    e0 = scatter_sum(
  File "/usr/local/lib/python3.10/dist-packages/mace/tools/scatter.py", line 36, in scatter_sum
    index = _broadcast(index, src, dim)
  File "/usr/local/lib/python3.10/dist-packages/mace/tools/scatter.py", line 18, in _broadcast
    if src.dim() == 1:
AttributeError: 'NoneType' object has no attribute 'dim'
```

### Steps to reproduce the problem

Load MACE checkpoint to an ase calculator and use it to get the atoms potential energy

```
from ase import Atoms, units
from ase.io import read, write
from matsciml.interfaces.ase import MatSciMLCalculator
from matsciml.datasets.transforms import (
    PeriodicPropertiesTransform,
    PointCloudToGraphTransform,
)

from ase import Atoms, units
from ase.md.verlet import VelocityVerlet

from matsciml.interfaces.ase import MatSciMLCalculator
from matsciml.datasets.transforms import (
    PeriodicPropertiesTransform,
    PointCloudToGraphTransform,
)

cp_path = "/mace/checkpoint/path"

d = 2.9
L = 10.0

atoms = Atoms("C", positions=[[0, L / 2, L / 2]], cell=[d, L, L], pbc=[1, 0, 0])

calc = MatSciMLCalculator.from_pretrained_force_regression(
    cp_path,
    transforms=[
        PeriodicPropertiesTransform(6.0, True),
        PointCloudToGraphTransform("pyg"),
    ],
)
atoms.calc = calc
[...]

PR #276, opened 2024-08-19
add missing data in format atoms

lory-w:
Related to #275 
Part of the problem is ase atoms is missing `ptr` and `batch` when converting to dictionary.

### DIFF
--- a/matsciml/interfaces/ase/base.py
+++ b/matsciml/interfaces/ase/base.py
@@ -212,6 +212,9 @@ class MatSciMLCalculator(Calculator):
         data_dict["pos"] = pos
         data_dict["atomic_numbers"] = atomic_numbers
         data_dict["cell"] = cell
+        # ptr and batch are usually expected by MACE even if it's a single graph
+        data_dict["ptr"] = torch.tensor([0])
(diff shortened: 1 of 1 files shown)


melo-gonzo:
A note for the future if any other errors pop up around this : we can also use `concatenate_keys` from `matsciml.datasets.utils` to add these attributes in a more robust way.

laserkelvin:
@melo-gonzo can you elaborate?

Just checking to make sure I understand what you mean

melo-gonzo:
I patched this in my mp-tests branch by just using that function as a final step before passing into `.predict` in the calculate function. Just wanted to note it in case we see any issues again
```python
    def calculate(
        self,
        atoms=None,
        properties: list[Literal["energy", "forces"]] = ["energy", "forces"],
        system_changes=...,
    ) -> None:
        # retrieve atoms even if not passed
        Calculator.calculate(self, atoms)
        # get into format ready for matsciml model
        data_dict = self._format_pipeline(atoms)
        # run the data structure through the model
        data_dict = concatenate_keys([data_dict])
        output = self.task_module.predict(data_dict)
        ...
```

Issue #290, opened 2024-09-16
[Feature request]: Support `*args` unpacking when creating classes in the experiment CLI


laserkelvin:
### Feature/behavior summary

Some classes unpack a list of arguments (`*args`) as part of the instantiation process, namely `matsciml.lightning.callbacks.LossScalingSchedule`. Essentially, what we want to be able to do in the `trainer` configuration is this:

```yaml
callbacks:
    - class_path: matsciml.lightning.callbacks.LossScalingScheduler
      init_args:
          - class_path: matsciml.lightning.loss_scaling.SigmoidScalingSchedule
            init_args:
                key: "energy"
                initial_value: 1.0
```

The idea is to allow `init_args` be a list of arguments, each of which are specs to create objects from; the API version would look like this:

```python
LossScalingScheduler(
    SigmoidScalingSchedule("energy", 1.0, ...)
)
```

Right now if the above configuration is used, we get the following error:

```python
TypeError: LossScalingScheduler.__init__() got an unexpected keyword argument 'class_path'
```

### Request attributes

- [X] Would this be a refactor of existing code?
- [ ] Does this proposal require new package dependencies?
- [ ] Would this change break backwards compatibility?
- [ ] Does this proposal include a new model?
- [ ] Does this proposal include a new dataset?
- [ ] Does this proposal include a new task/workflow?

### Solution description

Potentially modify `experiments.utils.utils.instantiate_arg_dict` to support unpacking a list of class specifications. Trouble is this might need a bit more complexity since you might need to actually inspect the arguments (i.e. with `argspec`).

PR #291, opened 2024-09-17
CLI `instantiate_arg_dict` Update

melo-gonzo:
This PR addresses the issue raised in #290 by passing nested classes and init_args as an *args style argument instead of kwargs.


laserkelvin (approved):
Tested offline and it's functional, thanks a lot for doing this!

PR #294, opened 2024-09-25
Parity inference task

laserkelvin:
This PR implements a `ParityInferenceTask` (with the lack of a better naming for it), which is used to run a datasplit through a model to evaluate predicted vs. actual, beyond just looking at reduced metrics like MSE/MAE. This task should be pretty straightforward to set up and run, and does not need additional logger configuration. After running `trainer.predict`, an `inference_data.json` is produced in the experiment folder (`trainer.log_dir/<name>/inference_data.json`) that can then be reviewed offline.

- Implements a `ParityData` structure which is mostly just a helper class for accumulating data.
- Implements the `ParityInferenceTask`, which in itself just provides the `predict_step` and `on_predict_epoch_end` functions to be called in the PyTorch Lightning `predict` pipeline.
- Unit test that verifies functionality on three datasets using EGNN. This tests simple functionality end-to-end, up to checking that the contents of the output JSON exists at all.
- Updated the documentation to include an "Inference" section, of which the first entry is this approach.

### DIFF
--- a/docs/source/inference.rst
+++ b/docs/source/inference.rst
@@ -36,7 +36,9 @@ The ``ParityInferenceTask`` helps perform this task by using the PyTorch Lightni
 The default ``Trainer`` settings will create a ``lightning_logs`` directory, followed by an experiment
 number. Within it, once your inference run completes, there will be a ``inference_data.json`` that you
 can then load in. The data is sorted by the name of the target (e.g. ``energy``, ``bandgap``), under
-these keys, ``predictions`` and ``targets``.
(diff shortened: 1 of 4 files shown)


melo-gonzo (approved):
Looks good and useful! A lot of failing tests, not sure what is relevant. Double check if these are applicable, then good to merge

laserkelvin:
I'm pretty certain the tests that fail are the usual suspects. I did pick up something that broke everything (fixed in 2faa68d), but otherwise seems to be fine. I'm going to merge!

PR #299, opened 2024-09-30
ASE Calculator Updates

melo-gonzo:
This pr updates the ASE calculator interface to accommodate a few changes:

* Use models that were not trained with matsciml. This eliminates the requirement for the model to be one of `ForceRegressionTask`, etc. 
* Map from model output keys to keys that are expected by ASE. Currently, the calculator manually renames model outputs to expected outputs by ase, for example the model output `force` gets mapped into `forces`. This mapping should now be passes as a dictionary to the calculator on instantiation if required. 
* The calculator will now iterate through the properties passed into `_calculate` to build the results dictionary.
* Add the `concatenate_keys` function which also adds properties to PyG graphs that are expected by models like MACE, add a few extra keys to `_format_atoms`, and change when type casting is called. 
* Updated tests to reflect new changes, added test using a pretrained `matgl` model.

### DIFF
--- a/matsciml/interfaces/ase/base.py
+++ b/matsciml/interfaces/ase/base.py
@@ -158,7 +158,7 @@ class MatSciMLCalculator(Calculator):
             specifies how model outputs should be mapped to Calculator expected
             results. for example {"ase_expected": "model_output"} -> {"forces": "force"}
         matsciml_model : bool, default True
-            flag indiciating whether model was trained with matsciml or not.
+            flag indicating whether model was trained with matsciml or not.
         """
         super().__init__(
             restart, label=label, atoms=atoms, directory=directory, **kwargs
--- a/matsciml/interfaces/ase/base.py
+++ b/matsciml/interfaces/ase/base.py
@@ -97,7 +97,6 @@ class MatSciMLCalculator(Calculator):
         directory=".",
         conversion_factor: float | dict[str, float] = 1.0,
         multitask_strategy: str | Callable | mt.AbstractStrategy = "AverageTasks",
-        data_type: torch.dtype | None = None,
         output_map: dict[str, str] | None = None,
         matsciml_model: bool = True,
         **kwargs,
@@ -150,10 +149,6 @@ class MatSciMLCalculator(Calculator):
             to ``ase``. If a single ``float`` is passed, we assume that
             the conversion is applied to the energy output. Each factor
             is multiplied with the result.
-        data_type : torch.dtype | None, default None
-            if specified, will convert data to this type instead
-            of looking at ``self.task_module.dtype`` which may not be
-            available in the case of 3rd party models.
         output_map : dict[str, str] | None, default None
             specifies how model outputs should be mapped to Calculator expected
             results. for example {"ase_expected": "model_output"} -> {"forces": "force"}
@@ -198,7 +193,6 @@ class MatSciMLCalculator(Calculator):
                 )
             multitask_strategy = cls_name()
         self.multitask_strategy = multitask_strategy
-        self.data_type = data_type
         self.matsciml_model = matsciml_model
         self.output_map = dict(
             zip(self.implemented_properties, self.implemented_properties)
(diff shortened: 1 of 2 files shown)


matsciml/interfaces/ase/base.py
```diff
@@ -212,9 +235,8 @@ def _format_atoms(self, atoms: Atoms) -> DataDict:
         data_dict["pos"] = pos
         data_dict["atomic_numbers"] = atomic_numbers
         data_dict["cell"] = cell
-        # ptr and batch are usually expected by MACE even if it's a single graph
-        data_dict["ptr"] = torch.tensor([0])
```
laserkelvin, line 216: Do you have to get rid of these?

matsciml/interfaces/ase/base.py
```diff
@@ -248,24 +266,33 @@ def calculate(
...
+            data_dict = recursive_type_cast(
+                data_dict,
+                self.dtype,
+                ignore_keys=["atomic_numbers"],
+                convert_numpy=True,
+            )
+            # run the data structure through the model
+            output = self.task_module.predict(data_dict)
+        else:
+            output = self.task_module.predict(atoms)
```
laserkelvin, line 284: So this requires other models to have a predict method as well?

If it's a model that _isn't_ a `matsciml` model, I don't know if it's a better idea to just simply run their `__call__` or `forward` method. If you don't think that's a good idea or have a reason for doing so, you should add in an assertion in the `__init__` that checks that the input model has a `predict` function.

matsciml/interfaces/ase/base.py
```diff
@@ -248,24 +266,33 @@ def calculate(
...
-                self.results["forces"] = output["force"].detach().numpy()
-            if "stress" in output:
-                self.results["stress"] = output["stress"].detach().numpy()
-            if "dipole" in output:
-                self.results["dipole"] = output["dipole"].detach().numpy()
+            # add outputs to self.results as expected by ase, as specified by ``properties``
+            # "ase_properties" are those in ``properties``.
+            for ase_property in properties:
+                model_output = output.get(self.output_map[ase_property], None)
+                if model_output is not None:
```
laserkelvin, line 294: Do you want to add an assertion/throw an error instead? If we expect the property to be calculated, then it should be present right?

matsciml/interfaces/ase/tests/test_ase_calc.py
```diff
@@ -62,8 +70,51 @@ def test_egnn_dynamics(egnn_config: dict, test_pbc: Atoms, pbc_transform: list):
...
+    matgl_model.forward = MethodType(forward, matgl_model)
+    matgl_model.predict = MethodType(forward, matgl_model)
+
+    calc = MatSciMLCalculator(matgl_model, matsciml_model=False)
+    pos = np.random.normal(0.0, 1.0, size=(10, 3))
+    # Using a different atoms object due to pretrained model atom embedding expecting
+    # a different range of atomic numbers.
+    atomic_numbers = np.random.randint(1, 94, size=(10,))
+    atoms = Atoms(numbers=atomic_numbers, positions=pos)
+    atoms = atoms.copy()
```
laserkelvin, line 115: Is there a reason you need to `copy()`?

laserkelvin (changes requested):

matsciml/interfaces/ase/base.py
melo-gonzo: Adding in `[concatenate_keys](def concatenate_keys()` adds these attributes, so they dont need to be added manually.

matsciml/interfaces/ase/base.py
melo-gonzo: Yeah, actually I think just running their forward method would be good. Updated here [81f1ab0](https://github.com/IntelLabs/matsciml/pull/299/commits/81f1ab0a32c490ddb9c3d8377b16ea01c8b50420)

matsciml/interfaces/ase/tests/test_ase_calc.py
melo-gonzo: Removed this bit here [81f1ab0](https://github.com/IntelLabs/matsciml/pull/299/commits/81f1ab0a32c490ddb9c3d8377b16ea01c8b50420).

matsciml/interfaces/ase/base.py
melo-gonzo: Added here [81f1ab0](https://github.com/IntelLabs/matsciml/pull/299/commits/81f1ab0a32c490ddb9c3d8377b16ea01c8b50420)

PR #300, opened 2024-10-01
Lightning Refactor: `pytorch_lightning` -> `lightning.pytorch`

melo-gonzo:
This pr migrated the pytorch lightning dependency from `pytorch_lightning` to `lightning` and replaces the imports. Since this pr touches a lot of files, there are some formatting changes, and just a few updates to tests, mostly to add a missing argument for MACE. 

Failing tests are mostly due to materials project trajectory dataset not having a devset, and some with SAM which have been failing in recent PR's. Also I did not fix any pre-commit related items dealing with `cdvae` things.


pyproject.toml
```diff
@@ -14,7 +14,7 @@ classifiers = [
 ]
 dependencies = [
   "psutil",
-  "pytorch_lightning==2.2.4",
+  "lightning==2.4.0",
```
laserkelvin, line 17: I would actually loosen this requirement - we're pinning to 2.4.0 because of `matgl` right? But technically, `matsciml` itself doesn't depend on 2.4.0 features. If it doesn't matter much to you, I think maybe setting it as `lightning>=2.4.0` might be a better way to go.

Philosophically I'm trying to relax these package versions a little bit to make it less nightmarish to put together the environment from scratch when needed.

laserkelvin (approved):
Functionally it looks good to me. Just left the one comment, and should be good to go from there.

matsciml/datasets/alexandria/devset/data_000.lmdb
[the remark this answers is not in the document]
melo-gonzo: I think they were regenerated when i ran tests locally. i reverted to the old files.

pyproject.toml
melo-gonzo: updated here [6351a98](https://github.com/IntelLabs/matsciml/pull/300/commits/6351a9897380ef5045f1f52c833538525fb751de)

laserkelvin:
LGTM, ready to merge

PR #302, opened 2024-10-04
Stress Calculation in `ForceRegressionTask`

melo-gonzo:
This PR adds the ability to perform stress calculations inside the `ForceRegressionTask`, and is inspired by the implementation here: https://github.com/ACEsuit/mace/blob/main/mace/modules/utils.py. This functionality if usefull in downstream tasks that require this parameter, and will be helpful for future MatSciML tasks such as a `StressRegressionTask` or similar.


laserkelvin:
I personally prefer `StressfulTask`

matsciml/models/base.py
```diff
@@ -1901,61 +1931,115 @@ def energy_and_force(
...
-            return energy, force
+
+            if self.compute_stress:
+                virials = outputs[1]
+                stress = torch.zeros_like(displacement)
+                cell = batch["cell"].view(-1, 3, 3)
+                volume = torch.linalg.det(cell).abs().unsqueeze(-1)
+                stress = virials / volume.view(-1, 1, 1)
+                stress = torch.where(
+                    torch.abs(stress) < 1e10, stress, torch.zeros_like(stress)
```
laserkelvin, line 1956: What is this hard coded value?

For something like this, I don't actually know if it's actually more performant to just clip values, instead of using `torch.where` where you have to allocate a `zeros_like` tensor.

matsciml/models/base.py
```diff
@@ -2178,6 +2262,43 @@ def _make_normalizers(self) -> dict[str, Normalizer]:
                 )
         return normalizers
 
+    @staticmethod
+    def get_symmetric_displacement(
```
laserkelvin, line 2266: If you were inspired by another repo for this function (and others), please add a link to them in the function's docstring

pyproject.toml
laserkelvin, line 1: How come you removed this?

laserkelvin (changes requested):

matsciml/models/base.py
melo-gonzo: added [d0be28d](https://github.com/IntelLabs/matsciml/pull/302/commits/d0be28d4874b06a68e917c0bc5a598dc548c82f4)

pyproject.toml
melo-gonzo: Not sure how that got purged. Updated [acd5c4a](https://github.com/IntelLabs/matsciml/pull/302/commits/acd5c4a393dca2dc930c7b77f8a0749f96677b71)

matsciml/models/base.py
melo-gonzo: this is an inspired bit from [here](https://github.com/ACEsuit/mace/blob/118a514efde34d963666118ce45360e94d648ef5/mace/modules/utils.py#L64). im assuming it is helpful in tasks which compute losses from stress. happy to remove the line and reconsider adding it back if/when we do stress related training.

matsciml/models/base.py
```diff
@@ -677,10 +677,9 @@ def __init__(
         encoder: nn.Module | None = None,
         encoder_class: type[nn.Module] | None = None,
         encoder_kwargs: dict[str, Any] | None = None,
-        loss_func: type[nn.Module]
-        | nn.Module
-        | dict[str, nn.Module | type[nn.Module]]
-        | None = None,
+        loss_func: (
+            type[nn.Module] | nn.Module | dict[str, nn.Module | type[nn.Module]] | None
```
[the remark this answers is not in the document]
melo-gonzo: This is just formatting

melo-gonzo:
> For now this is fine, but I have to say that `ForceRegressionTask` is now reaching unmaintainable levels, now that we have both frame averaging and stress/virial logic spread out across `process_embedding` and other functions. We really need to think about reabstracting stuff where it makes sense. Not saying we need to do it for this PR, but sooner rather than later.

I agree, it could use some organizing.

PR #303, opened 2024-10-07
Quantile-based loss mechanism

laserkelvin:
This PR implements a dynamic quantile-based loss, which computes and rescales specified loss functions depending on quantiles bins within a batch. An accompanying suite of tests will test argument capabilities as well as the end-to-end training with a nominal `EGNN` and `ForceRegressionTask`.

How the loss function works is detailed in the docstring. At a high level, we take the targets and compute their respective quantiles, and use these values to bin the targets as to allow the user to define which ranges to (de)emphasize.


melo-gonzo (approved):
looks nice!

Issue #267, opened 2024-08-06
[Feature request]: Refactor `MaterialsProjectDataset` to not serialize `pymatgen` `Structures` in LMDB


laserkelvin:
### Feature/behavior summary

Currently, the workflow implemented for `MaterialsProjectDataset` will save and reload a `pymatgen.Structure` object. The issue with this is that it is very intimately tied to the version of `pymatgen`, where small API changes can make it difficult to reload the dataset in later versions.

### Request attributes

- [X] Would this be a refactor of existing code?
- [ ] Does this proposal require new package dependencies?
- [X] Would this change break backwards compatibility?
- [ ] Does this proposal include a new model?
- [ ] Does this proposal include a new dataset?
- [ ] Does this proposal include a new task/workflow?

### Solution description

If we can refactor it so that `Structure`s are created at load time - in line with other dataset implementations - it would make it break this dependency...breaking.

We would have to re-process the existing LMDBs being distributed, and make sure that the data is stored as just plain coordinates, atoms, and lattice parameters.

PR #305, opened 2024-10-10
`pymatgen` structure creation patch

### LINKED ISSUE #304: [Bug]: `Lattice` objects created are inconsistent with existing ones
### Expected behavior

The periodic properties calculation workflow currently creates a new `pymatgen.core.Structure` object, regardless of whether it exists or not, and based either on an existing lattice matrix or from parameters.

When using `MaterialsProjectDataset` and children, @melo-gonzo pointed out that the serialized structures are not identical to the ones produced by the `matsciml` workflow, which is **not** what we would expect as it then leads to a mismatch in input structures and labels.

### Actual behavior

In a set of notebook tests, the image below shows the creation of a `Lattice` object from parameters, versus passing an existing `Structure`'s lattice matrix directly: note the differences in the resulting matrix.

Screenshot from 2024-10-09 15-06-04

If you inspect the resulting structure's coordinates, nothing immediately stands out but if you visualize the structure, they are _very_ different. This is a huge issue for force prediction, as the atoms are switched: the top panel below shows the `matsciml` emitted structure, and the lower panel is from the serialized structure.

Screenshot from 2024-10-09 15-10-03

As it turns out, at some point, `pymatgen` introduced a `vesta` argument to `Lattice.from_parameters`. The default value is `False`, but if set to `True`, the generated and serialized structures are then consistent.

Code snippets shown above; compare serialized structure with that produced by the `PeriodicPropertiesTransform` workflow.

### Specifications

Mainly relevant here is `pymatgen==2023.9.25`

laserkelvin:
This PR resolves #304 by refactoring `PeriodicPropertiesTransform` in two ways:

1. We will reuse a serialized `structure` if it is interpreted as a `pymatgen.core.Structure` for the periodic boundary neighbors calculation, and return directly.
2. If the above condition fails, we'll continue down the old path, but `Lattice.from_parameters` is now passed `vesta=True` in all cases if creating from lattice parameters. This set up to happen in both the transform, as well as the subroutine `matsciml.datasets.utils.make_pymatgen_periodic_structure`.

The impact of this change is likely only with datasets that do not have a `cell` attribute in their data. e.g. `LiPSDataset` and Open Catalyst expose `cell` which is used for the `Lattice` creation, and if the fully qualified lattice matrix is given then the resulting lattice is consistent. Only when no `cell` is available is when `Lattice.from_parameters` is called.

A related issue is #267, if and when we get to it. There would not be inconsistencies if the `pymatgen.core.Structure` is created from first principles, as opposed to loading a serialized one.

### DIFF
--- a/matsciml/datasets/transforms/pbc.py
+++ b/matsciml/datasets/transforms/pbc.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import torch
-from pymatgen.core import Lattice
+from pymatgen.core import Lattice, Structure
 
 from matsciml.common.types import DataDict
 from matsciml.datasets.transforms.base import AbstractDataTransform
@@ -37,6 +37,12 @@ class PeriodicPropertiesTransform(AbstractDataTransform):
     def __call__(self, data: DataDict) -> DataDict:
         for key in ["atomic_numbers", "pos"]:
             assert key in data, f"{key} missing from data sample!"
+        if "structure" in data:
+            structure = data["structure"]
+            if isinstance(structure, Structure):
+                graph_props = calculate_periodic_shifts(
+                    structure, self.cutoff_radius, self.adaptive_cutoff
+                )
         if "cell" in data:
             # squeeze is used to make sure we remove empty dims
             lattice = Lattice(data["cell"].squeeze())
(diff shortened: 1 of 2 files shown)


melo-gonzo (approved):
Thanks for the updates. Glad this is all sorted.

PR #306, opened 2024-10-10
Relaxed `pyproject.toml` specifications and improved dependency resolution

laserkelvin:
This PR is intended to significantly simplify the `matsciml` installation process.

1. We provide the exact `pip` command to install all dependencies, which provide the necessary binary distribution links for both PyG and DGL.
2. Included refreshed `conda.yml` specification, which should now give the equivalent process as `pip` but also wrapping it in a managed environment.
3. We relax the `dgl` requirement, which AFAIK does not have a strict requirement anymore, and by unpinning it makes it less likely to run into issues to missing GraphBolt, etc.
4. Updated `matsciml.common.package` to be a little more specific about the package dependencies that fail most frequently, i.e. DGL and PyG. We will actually try and load in these packages and report back any exceptions, rather than just enabling them in the package registry if they're installed. This helps with debugging because rather than failing when trying to run a model, it will fail much earlier as it will indicate a botched installation/environment.
5. Related to the above, introduced `loguru` logging which will just make things look more tidy.

The main effect of this update is to really streamline the installation process: the README now includes copy-pasteable commands that should just work out of the box. In order to do so, I've had to take the executive decision to bump PyTorch to 2.4 or 2.3, primarily because we need those determined and hardcoded for the URLs to work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,22 +44,36 @@ The examples outlined in the next section how to get started with Open MatSci ML
 
 - `Docker`: We provide a Dockerfile inside the `docker` that can be run to install a container using standard docker commands.
 - `mamba`: We have included a `mamba` specification that provides a complete out-of-the-box installation. Run `mamba env create -n matsciml --file conda.yml`, and will install all dependencies and `matsciml` as an editable install.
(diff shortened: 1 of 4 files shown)


melo-gonzo (approved):
looks great!

PR #310, opened 2024-10-21
ASE Backend For PBC Computation

melo-gonzo:
This PR adds a flag to the `PeriodicBoundaryCondition` transform to use either `pymatgen` or `ase` as a backend. The ASE backend is implemented and makes use of the [NeighborList](https://wiki.fysik.dtu.dk/ase/_modules/ase/neighborlist.html#NeighborList) utility in ASE. This requires creating an atoms object, and specifying a cutoff distance per node. The two functions `calculate_ase_periodic_shifts` and `calculate_periodic_shifts` are quite similar, however hard to combine due to the dependency on the different backends, so they are kept separate for now. 

Some parameters that may have an impact on results:
`NeighborList`: `self_interaction`, `bothways`
`Atoms`: `pbc`

For example, blindly applying PBC in x, y, and z for a structure which is not meant to be periodic in these directions may impact results.

### DIFF
--- a/matsciml/datasets/transforms/pbc.py
+++ b/matsciml/datasets/transforms/pbc.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+from typing import Literal
+
 import numpy as np
 import torch
 from pymatgen.core import Lattice, Structure
@@ -35,7 +37,7 @@ class PeriodicPropertiesTransform(AbstractDataTransform):
         self,
         cutoff_radius: float,
         adaptive_cutoff: bool = False,
-        backend: str = "pymatgen",
+        backend: Literal["pymatgen", "ase"] = "pymatgen",
     ) -> None:
         super().__init__()
         self.cutoff_radius = cutoff_radius
--- a/matsciml/datasets/transforms/pbc.py
+++ b/matsciml/datasets/transforms/pbc.py
@@ -1,13 +1,14 @@
 from __future__ import annotations
 
-import torch
 import numpy as np
+import torch
 from pymatgen.core import Lattice, Structure
 
 from matsciml.common.types import DataDict
 from matsciml.datasets.transforms.base import AbstractDataTransform
 from matsciml.datasets.utils import (
     calculate_periodic_shifts,
+    calculate_ase_periodic_shifts,
     make_pymatgen_periodic_structure,
 )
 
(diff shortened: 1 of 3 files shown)


matsciml/datasets/transforms/pbc.py
```diff
@@ -30,10 +31,16 @@ class PeriodicPropertiesTransform(AbstractDataTransform):
     a large cut off for the entire dataset.
     """
 
-    def __init__(self, cutoff_radius: float, adaptive_cutoff: bool = False) -> None:
+    def __init__(
+        self,
+        cutoff_radius: float,
+        adaptive_cutoff: bool = False,
+        backend: str = "pymatgen",
```
laserkelvin, line 38: FYI you can type "choices" as `Literal["pymatgen", "ase"]` to hint at  what is expected

matsciml/datasets/utils.py
```diff
@@ -751,3 +754,69 @@ def _all_sites_have_neighbors(neighbors):
...
+
+
+def calculate_ase_periodic_shifts(data, cutoff_radius, adaptive_cutoff):
+    cell = data["cell"]
+
+    atoms = ase.Atoms(
+        positions=data["pos"],
+        numbers=data["atomic_numbers"],
+        cell=cell.squeeze(0),
+        pbc=(True, True, True),
```
laserkelvin, line 766: Add a comment here to make sure we're hardcoding this for now

laserkelvin (commented):
Added minor comments, but the I might have a bigger lift/ask: could you write a unit test to ensure the products (i.e. offsets, `src` and `dst` nodes) are the same?

You could potentially do a really high level comparison: make `torch_geometric.data.Data` with the edges from both algorithms, pack the data into the graphs, and just do a `pmg_graph == ase_graph` assertion.

laserkelvin:
Related to #267 - possibly leading up to replacing `pymatgen` with equivalent `ase` functionality.

matsciml/datasets/transforms/pbc.py
melo-gonzo: updated [9e6919b](https://github.com/IntelLabs/matsciml/pull/310/commits/9e6919bef2b6d2ac8652cc40130cfe87e18bed06)

matsciml/datasets/utils.py
melo-gonzo: updated [4d5eb4c](https://github.com/IntelLabs/matsciml/pull/310/commits/4d5eb4c6010d3b788d76f4bb5d33f513dd336c42). FYI it seems that pymatgen structures are also "3d periodic" so i think this is fine for now.

melo-gonzo:
I also added two tests to check for equivalence of src and dst node calculation [here](https://github.com/IntelLabs/matsciml/pull/310/commits/a0af875f2042caa0b9bb9b0e511d9069f6aca351)

laserkelvin (approved):
LGTM, thanks for doing this!

PR #314, opened 2024-11-08
Scatter add refactor

### LINKED ISSUE #313: [Feature request]: Refactor `FAENet` to not require `torch_scatter`
### Feature/behavior summary

`FAENet` uses a single `torch_scatter.scatter` call for the readout, which in principle could be refactored to now use `torch.scatter_add` and remove this hard dependency. The advantage of doing so is that hardware that do not have `torch_scatter` kernels implemented can still use `FAENet`.

We would have to test it numerically to make sure it's producing the same result.

### Request attributes

### Related issues

_No response_

### Solution description

Rewrite that block to use `scatter_add` or something equivalent. `MACE` implements its own `scatter` operation, which could be re-used globally by all models.

### Additional notes

_No response_

laserkelvin:
This PR closes #313 by refactoring the code base to use the `scatter_sum` ported over by MACE. This allows `FAENet` to be used without the hard `torch_scatter` dependency, along with `SchNet`.

### DIFF
--- a/matsciml/models/pyg/__init__.py
+++ b/matsciml/models/pyg/__init__.py
@@ -19,6 +19,7 @@ else:
 # load models if we have PyG installed
 if _has_pyg:
     from matsciml.models.pyg.egnn import EGNN
+    from matsciml.models.pyg.faenet import FAENet
     from matsciml.models.pyg.mace import MACE, ScaleShiftMACE
 
     __all__ = ["EGNN", "FAENet", "MACE", "ScaleShiftMACE"]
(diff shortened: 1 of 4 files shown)


melo-gonzo:
Should we also purge from [here](https://github.com/IntelLabs/matsciml/blob/86c626151b595fca5a84bec52eaa4ccd61b9be39/matsciml/models/pyg/mace/tools/scatter.py) as part of the PR?

laserkelvin:
We can - but I felt like I wanted to purge the unused code alongside it too in the rest of the module, and just reduce it down to the wrapper.

melo-gonzo (approved):

PR #315, opened 2024-11-08
Structured model output type

laserkelvin:
This PR introduces `pydantic` based data structures for homogenizing model outputs.

1. I've refactored the `Embeddings` class to use `pydantic`, away from the native `dataclass`. There wasn't much modification otherwise, as there aren't as many validation checks to be done with it.
2. The new `ModelOutput` data structure should be the way forward for all models; refactoring that can come later. The main advantage of this class is to homogenize and validate model outputs, and provide meaningful/helpful ways to debug when issues arise (e.g. shape mismatches, batch size with energy predictions, etc.).

At a high level, `pydantic` will run type checking on the inputs (casting when possible), and run user-defined `before` and `after` checks at the `Field` (specific attributes) and `Model` levels - we use the whole gambit for checking. The included unit tests should demonstrate how the class is expected to behave: most keyword arguments are optional (an upcoming PR will explain why), and we will perform checks if and when the relevant data is available.

Keep in mind `pydantic` models _only_ accept kwargs for creation - i.e. we have to be a bit more verbose/explicit, unfortunately.

### DIFF
--- a/matsciml/common/tests/test_types.py
+++ b/matsciml/common/tests/test_types.py
@@ -22,3 +22,20 @@ def test_incorrect_force_shape():
     """This passes a force tensor with too many dimensions"""
     with pytest.raises(ValidationError):
         types.ModelOutput(batch_size=8, forces=torch.rand(32, 4, 3))
+
+
+def test_consistency_check_pass():
+    types.ModelOutput(
+        batch_size=8, forces=torch.rand(32, 3), node_energies=torch.rand(32, 1)
+    )
+
+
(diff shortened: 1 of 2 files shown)


laserkelvin:
I reverted the `Embeddings` change - it broke the end-to-end EGNN test and reminded me that every instance of `Embeddings` being used doesn't follow the strict `pydantic` kwarg way. It wasn't necessary anyway :D

melo-gonzo (approved):

Issue #104, opened 2024-01-20
[Feature request]: Homogenization of data structures and physical representations


laserkelvin:
### Feature/behavior summary

To ensure consistency in modeling, each dataset in Open MatSciML Toolkit should have uniform (or near uniform) *kinds* of data. For example, whether coordinates provided are fractional or Cartesian, ensuring every dataset has sufficient information to represent each data sample in a physically meaningful way, such as periodic boundary conditions (for use in e.g. shift vectors).

### Request attributes

- [X] Would this be a refactor of existing code?
- [ ] Does this proposal require new package dependencies?
- [ ] Would this change break backwards compatibility?
- [ ] Does this proposal include a new model?
- [ ] Does this proposal include a new dataset?
- [ ] Does this proposal include a new task/workflow?

### Solution description

A good place to start would be to make sure each `devset`, and subsequently any serialized datasets we have conform to the following:

1. Check if the coordinates are fractional or not (if there are values outside of 0 and 1 then they're likely Cartesian)
2. Check to make sure we have enough information to create a Lattice object, can be just a cell key, or have the lattice parameters like materials project
3. Generally just print and list out the keys in the sample, construct a table of them, so that we can help contribute to #97

We should also check other projects, like Colabfit, to see what extent we can try and conform to community standards, too.

### Additional notes

Can't assign Bin yet, but would be good for Bin to aggregate information, and between him and @melo-gonzo to help craft PRs to address things after the survey is done.

bmuaz:
I had the same thoughts about the data structures and will be happy to work on it with Carmelo.

PR #316, opened 2024-11-12
Introducing structured model outputs

laserkelvin:
This PR is semi-related to #104, allowing flexibility for wrappers and full implementations to potentially have a common output format that is also "backwards" compatible. It depends on `ModelOutput` from #315, so please review and merge that before this PR.

- Added a private attribute of the base `AbstractTask` class (which graph models inherit) called `__skip_output_heads__`, which `BaseTaskLitModules` will check this to determine if output heads are needed or not. If we do not use them, we do not initialize the `OutputHeads` to save on parameters/ensure we don't have unused parameters.
- I have refactored the `MACEWrapper` as an example of how to do this, and we should do so for other wrappers as well (e.g. `CHGNet`). I've also updated the `sphinx` documentation to detail how this can be used.
- I've ensured that the regular workflow (with PyG EGNN) isn't broken, and included unit tests for `MACEWrapper` to run the forward pass and an end-to-end with `ForceRegressionTask`.

### DIFF
--- a/matsciml/models/base.py
+++ b/matsciml/models/base.py
@@ -904,19 +904,24 @@ class BaseTaskModule(pl.LightningModule):
         batch: dict[str, torch.Tensor | dgl.DGLGraph | dict[str, torch.Tensor]],
     ) -> dict[str, torch.Tensor]:
         encoder_outputs = self.encoder(batch)
-        if not isinstance(encoder_outputs, (Embeddings, dict)):
-            raise RuntimeError(
-                f"Encoder model must emit a dict or `Embeddings` object. Got {encoder_outputs} instead."
-            )
-        if isinstance(encoder_outputs, Embeddings):
-            batch["embeddings"] = encoder_outputs
-            outputs = self.process_embedding(encoder_outputs)
+        # in the case that the model does not produce its own outputs,
+        # we will pass them through the output heads.
+        if not self.model.__skip_output_heads__:
+            if not isinstance(encoder_outputs, (Embeddings, dict)):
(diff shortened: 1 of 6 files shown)


laserkelvin:
CI passed on relevant things

matsciml/models/base.py
```diff
@@ -1871,10 +1922,38 @@ def forward(
...
+                encoder_outputs = self.encoder(batch)
+                if not isinstance(encoder_outputs, (Embeddings, dict, ModelOutput)):
+                    raise RuntimeError(
+                        f"Encoder model must emit a dict, `ModelOutput`, or `Embeddings` object. Got {encoder_outputs} instead."
+                    )
+                # sets the embeddings variable
+                if isinstance(encoder_outputs, Embeddings):
+                    embeddings = encoder_outputs
+                # for BYO output head cases
+                elif isinstance(encoder_outputs, ModelOutput):
```
melo-gonzo, line 1938: A comment that may not be super important to consider right at this moment but should be noted - some workflows, such as serving models with OpenKIM's `kusp` to then run benchmarks, require that node energies are present. Some models (such as mace) can output these directly and they may be worth hanging onto as well.

melo-gonzo (approved):
Looks good overall! Will be a nice improvement to have. Left one general comment about downstream workflows.

matsciml/models/base.py
laserkelvin: I think they're included in `ModelOutput` - in the wrapper output they're stashed as `node_energies`. That's what you mean right?

PR #318, opened 2024-11-15
Specifiable options for periodic neighbors calculations

### LINKED ISSUE #317: [Feature request]: Customization of periodic edge calculation
### Feature/behavior summary

Would be good to add new options for `PeriodicPropertiesTransform`, allowing the user to:

- [ ] Optionally remove self-loops
- [ ] Request the ensuing graph to be directed or not
- [ ] Specify a maximum number of edges per node; we can truncate based on sorted distances

### Request attributes

### Related issues

_No response_

### Solution description

This would require a set of refactors that could be bundled in a single PR or split up. The modifications could ideally be made _after_ running the calculation with either `ase` or `pymatgen`, and just modify edge resulting `src_nodes` and `dst_nodes` lists.

### Additional notes

_No response_

laserkelvin:
This PR closes #317 by allowing optional keyword arguments to the `PeriodicPropertiesTransform` workflow:

- `max_neighbors` provides a hard cap on the **destination** nodes for any atom. A counter is updated within the per-site loop, and will break out if we exceed the `max_neighbors` value.
- `allow_self_loops`, when set to `False` (which is now the default), will filter out self-loops as the last step, just before we return the updated data dictionary.

I also caught and fixed two bugs in the `ase` implementation:

- Using the right variable check for `adapative_cutoff`
- Missing `cell` reshape before passing it into the `einsum`

### DIFF
--- a/matsciml/datasets/utils.py
+++ b/matsciml/datasets/utils.py
@@ -868,6 +868,7 @@ def calculate_ase_periodic_shifts(
         "pos": coords,
     }
 
+    cell = rearrange(cell, "i j -> () i j")
     return_dict["offsets"] = einsum(return_dict["images"], cell, "v i, n i j -> v j")
     src, dst = return_dict["src_nodes"], return_dict["dst_nodes"]
     return_dict["unit_offsets"] = (
--- a/matsciml/datasets/utils.py
+++ b/matsciml/datasets/utils.py
@@ -721,6 +721,7 @@ def calculate_periodic_shifts(
     if np.any(structure.frac_coords > 1.0):
         logger.warning(
             f"Structure has fractional coordinates greater than 1! Check structure:\n{structure}"
+            f"\n fractional coordinates: {structure.frac_coords}"
         )
 
     def _all_sites_have_neighbors(neighbors):
--- a/matsciml/datasets/utils.py
+++ b/matsciml/datasets/utils.py
@@ -658,7 +658,10 @@ def make_pymatgen_periodic_structure(
 
 
 def calculate_periodic_shifts(
-    structure: Structure, cutoff: float, adaptive_cutoff: bool = False
+    structure: Structure,
+    cutoff: float,
+    adaptive_cutoff: bool = False,
+    max_neighbors: int = 1000,
 ) -> dict[str, torch.Tensor]:
     """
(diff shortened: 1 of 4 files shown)


matsciml/datasets/utils.py
```diff
@@ -724,11 +745,16 @@ def _all_sites_have_neighbors(neighbors):
 
     all_src, all_dst, all_images = [], [], []
     for src_idx, dst_sites in enumerate(neighbors):
+        site_count = 0
         for site in dst_sites:
+            if site_count > max_neighbors:
```
melo-gonzo, line 751: are `neighbors` sorted in any way? if not it might make sense to prune the lists after the loop.

matsciml/datasets/utils.py
```diff
@@ -815,9 +882,141 @@ def _all_sites_have_neighbors(neighbors):
...
+        return 1 / np.sin(x)
+
+    # convert to radians if angles are passed as degrees
+    if angles_are_degrees:
+        alpha = alpha * np.pi / 180.0
+        beta = beta * np.pi / 180.0
+        gamma = gamma * np.pi / 180.0
+
+    # this matrix is normally for fractional to cart
+    rotation = torch.tensor(
```
melo-gonzo, line 966: maybe add a link to the wiki where this came from, for future reference.

melo-gonzo (approved):
Look good overall, left two general comments

matsciml/datasets/utils.py
laserkelvin: I haven't fully verified it, but under the hood for `ase.neighborlist`, the [code](https://wiki.fysik.dtu.dk/ase/_modules/ase/neighborlist.html#PrimitiveNeighborList) uses `scipy.spatial.cKDTree` which should be sorted based on nearest neighbors: i.e. I think if it's not sorted for the whole list of indices, it should be at least for the same level of the tree.

In other words with a `max_neighbors` of 50+ you're definitely going to be getting the nearest neighbors, but towards the end of the 50, *those* neighbors might not be necessarily sorted.

matsciml/datasets/utils.py
laserkelvin: Added in 973a1b9

PR #319, opened 2024-11-18
`MACEWrapper` interface changes

laserkelvin:
This PR changes how `atomic_energies` are treated in the wrapper, supporting `dict` mappings and generally better documentation in the docstring to describe what arguments are used where.


melo-gonzo (approved):
LGTM

PR #333, opened 2024-12-17
Final edge creation

laserkelvin:
This PR modifies the `Edge` class to get exact parity in the number of edges with `ovito`, thanks to code contributed by Sajid Mannan from IIT Delhi.

We introduce an `is_undirected` flag to the `Edge` class that modifies the behavior of the `__eq__` for this class, which propagates to the behavior of how edges are treated in `set()` for determining uniqueness.

I've added a documented set of tests to explain what the expected behavior of the edge sets are.


melo-gonzo (approved):
LGTM

PR #335, opened 2024-12-20
(Pen)ultimate edge logic fix

laserkelvin:
This PR fixes the edge counts in the directed case (i.e. with symmetric edges) being overly aggressively filtered out.

I've refactored where the `set` based filtering of edges is done: instead of in the `Edge` class itself, we simply skip over all of the `set` loops if `is_undirected=False`. In other words, if `is_undirected=False`, we simply return the edges out of `ase` or `pymatgen` neighbor lists.

### DIFF
--- a/matsciml/datasets/utils.py
+++ b/matsciml/datasets/utils.py
@@ -963,24 +963,28 @@ def calculate_ase_periodic_shifts(
     )
     # not really needed but good sanity check
     assert np.all(distances <= cutoff_radius)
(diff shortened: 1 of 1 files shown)


melo-gonzo (approved):

laserkelvin:
With this patch, when `is_undirected` is set to False, we bypass the `max_neighbors` logic entirely. I've opened #336 to track.

PR #348, opened 2025-03-19
Refactor `DataSampleSchema` to use tensors

laserkelvin:
This PR refactors the `DataSampleSchema` to use `Annotated` abstraction instead of a `field` and `model` validators, and replacing the shape checking with methods in the new `matsciml.datasets.validators` module.

The intention behind these changes are:

1. Ensure that tensors are used first and foremost, which otherwise introduces barriers to usage in the full pipeline because tensors need to be moved to devices, expected dtypes, etc.
2. The tensor annotations in the model schema are now more semantically meaningful; e.g. `CoordTensor` has more implications to it than `NDArray['*, 3']`. For the user/developer, these effectively "named" tensors in the annotations are likely better to work with.
3. Reusable validator workflows: nominally they can be used with batched data as well.

### DIFF
--- a/matsciml/datasets/schema.py
+++ b/matsciml/datasets/schema.py
@@ -811,7 +811,7 @@ class DataSampleSchema(MatsciMLSchema):
             self.frac_coords = cart_frac_conversion(
                 self.cart_coords, *self.lattice_parameters, to_fractional=True
             )
-        if isinstance(self.frac_coords, NDArray):
+        if isinstance(self.frac_coords, torch.Tensor):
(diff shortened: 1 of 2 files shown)


smiret-intel (approved):
Looks like a good validation setup for commonly used Tensors. Can you add some brief documentation on the types for future users?

laserkelvin:
> Looks like a good validation setup for commonly used Tensors. Can you add some brief documentation on the types for future users?

They're not in the main pipeline just yet, but I've already started documenting them in the [readthedocs](https://matsciml.readthedocs.io/en/latest/schema.html)

PR #349, opened 2025-03-20
Static batch schema

laserkelvin:
This PR addresses an issue with the existing `BatchSchema` definition, which is done on-the-fly based on what information was actually packed into the individual data samples. The issue with this is that serialization is not possible, which prevents multiple data loader workers being used, which hinders training.

This is solved by creating a static definition of a `BatchSchema` that subclasses `DataSampleSchema`. The scope of testing is on a new HDF5 dataset with PyG graphs, and is functional for multi-GPU training.

Summary of changes:
- Defined `BatchSchema`, subclassing `DataSampleSchema`
- Added a `to` method for `DataSampleSchema` (which is inherited by `BatchSchema`) for data transfers to accelerators
- Added `transfer_batch_to_device` methods to `LightningModule` definitions, which facilitates the data transfer before model calls
- Unit tests for batching with the new `BatchSchema`

### DIFF
--- a/matsciml/datasets/generic.py
+++ b/matsciml/datasets/generic.py
@@ -11,11 +11,7 @@ import h5py
 from torch.utils.data import DataLoader, Dataset
 from lightning import pytorch as pl
 
-from matsciml.datasets.schema import (
-    DatasetSchema,
-    DataSampleSchema,
-    collate_samples_into_batch_schema,
-)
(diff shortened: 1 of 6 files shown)


smiret-intel (approved):
Does this imply we need to change the underlying data format to HDF5? I assume this works for native PyTorch as well.

